PR #1, opened 2021-04-10
Add MicroFPGA

jdeschamps:
Hi there,

Here is the device adapter for a FPGA board that the Ries lab (EMBL) and collaborators have been using for a few years on their microscopes. The platform is called [MicroFPGA](https://github.com/jdeschamps/MicroFPGA) and is based on an hobby FPGA board from Alchitry (Au or Cu, now produced by [SparkFun](https://www.sparkfun.com/categories/tags/alchitry)).

MicroFPGA generats various control signals used on microscopes such as a flexible laser triggering system, servomotors, TTL or PWM (used for instance together with a low-pass for analog output). Together with one of the FPGA board (Au), it can also read in analog signals.

We put some effort in making the [project repository](https://github.com/jdeschamps/MicroFPGA) as documented as possible so that it is easy for people to reproduce and install on their own microscope.

### DIFF
--- a/DeviceAdapters/MicroFPGA/Makefile.am
+++ b/DeviceAdapters/MicroFPGA/Makefile.am
@@ -1,0 +1,6 @@
+AM_CXXFLAGS = $(MMDEVAPI_CXXFLAGS)
+deviceadapter_LTLIBRARIES = libmmgr_dal_MicroFPGA.la
+libmmgr_dal_MicroFPGA_la_SOURCES = MicroFPGA.cpp MicroFPGA.h \
+   ../../MMDevice/MMDevice.h ../../MMDevice/DeviceBase.h
+libmmgr_dal_MicroFPGA_la_LIBADD = $(MMDEVAPI_LIBADD)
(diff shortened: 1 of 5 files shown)


nanthony21:
Thanks! Other than the minor thing that I commented on I think this is good to go.

jdeschamps:
Cheers! Where is the comment? I must have missed it. :)

nanthony21:
Ok, I was just asking to have the lines that are like: `#include "../../MMDevice/header_file.h"` changed to `#include "header_file.h"`.

The reason is just that the `MMDevice` folder is already added to the include paths in the project. If, in the future, the relative path to `MMDevice` changes then we will only need to adjust the path in the project property rather than changing it in every source file.

I made this change in all of the existing device adapters when migrating to this repository so I'd like to maintain that for new device adapters as well.

jdeschamps:
Makes sense! I implemented the changes.

nanthony21:
Looks good. This should hopefully show up in tomorrow's build.

PR #6, opened 2021-04-22
Fix SDK include paths 

nanthony21:
@nicost , I don't currently have the ability to test the build but this should fix the issue with the Spinnaker camera. The SDK include paths were accidentally reverted when migrating to this submodule.

Don't merge this yet as I am still checking all the other device adapters for similar mistakes. Since we changed all "3rdParty" include paths from `..\..\..\3rdParty` to `$(MM_3RDPARTYPRIVATE)`, Git recognizes them all as having changed which means they must be checked manually to see if the path is actually different.

So far I have noticed a few other potential issues which I'm not sure how to properly resolve:
1: The ABS device adapter code from before merging the submodule uses `#ifdef LINUX`, while the current code uses `#ifdef __linux__` I'm not sure which is correct.
2: The AndorSDK3 from before the merge uses `#if defined(__linux__)` while the current code uses `#if defined(linux)`
3: A change in the Basler device adapter has been committed to SVN in the time after we moved to using the submodule so the Git and SVN versions are now different. I really think we need to stop development on the SVN version.

### DIFF
--- a/DeviceAdapters/AndorSDK3/AndorSDK3.cpp
+++ b/DeviceAdapters/AndorSDK3/AndorSDK3.cpp
@@ -1531,7 +1531,7 @@ void CAndorSDK3Camera::RestartLiveAcquisition()
 
 AT_64 CAndorSDK3Camera::GetTimeStamp(unsigned char* pBuf)
 {
-#if defined(linux) && defined(_LP64)
+#if defined(__linux__) && defined(_LP64)
    typedef unsigned int    AT_U32;
 #else
    typedef unsigned long   AT_U32;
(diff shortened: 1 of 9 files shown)


nanthony21:
One further change which is not accidental but is worth noting is that any device adapters that ran custom post-build scripts have had those scripts removed. There is no reason for us want individual device adapters calling things like this:
```
    <PostBuildEvent>
      <Command>copy /Y "$(TargetDir)$(TargetFileName)" "C:\Program Files (x86)\Micro-Manager-1.4"</Command>
    </PostBuildEvent>
```

nanthony21:
issues 1 and 2 have been resolved. 3 has not.

nanthony21:
I'm unable to get PVCam to build, however this is also true if I check out a version of the code from before using this submodule.

nicost:
Ravi (maintaining the Basler adapter) will both push to svn and send us a PR soon.  Stopping svn access to the 1.4 repository means stopping development on the 1.4 branch altogether.  Companies like ASI have code very important to their customers that has not been ported to 2.0 and it will be a big project to do so.  I do not want preclude anyone from continuing to push changes to the 1.4 device adapters, and since we have these builds working, why not continue to let them run?

I do intend to check in now and then on who is pushing changes to 1.4, and nudge them to send us a PR as well.

We also have not solved the issue of SecretDeviceAdapters yet.  Some developers there have no other way but using svn to send us changes securely.

nanthony21:
I see. I plan to set up a better Git situation for the secret device adapters once I get to that part of the Meson build system. For now there is a branch at https://github.com/micro-manager/SecretDeviceAdapters called `svnSync` which is set to track the correct folder of the micromanager2 SVN repo. We can manually update it and merge it to `main` as needed.

PR #12, opened 2021-05-19
Custom streaming to disk from PVCAM adapter.

tomhanak:
This PR basically synchronizes changes from PVCAM adapter in Micro-Manager SVN to git.

The custom solution saving raw data to disk has been chosen to work around limited data throughput between C++ and Java code. It is needed mainly to stream data with Kinetix camera in speed mode (almost 5 GB/s). It is fully controlled via new parameters added to Parameter Browser.

The usage yet to be described on MM wiki.


nanthony21:
Thanks, looks good!

Issue #25, opened 2021-07-22
`setConfig` should always trigger an `onConfigGroupChanged` or a new signal should be created


ianhi:
# Problem

I would expect the `MMEventCallback`'s `onConfigGropChanged` method to be always be called whenever `MMCore:SetConfig` has been called. However, this is not the case as the only place that callback is called from is here:

https://github.com/micro-manager/mmCoreAndDevices/blob/0e5ba3c139ad5fab40aeec58bdd57c91f3c9c474/MMCore/CoreCallback.cpp#L496-L506

This means that if you follow the demo config and have a group of a channel with only a filter cube in it then any callbacks you have set up will not fire.

Here is an example using `pymmcore`:

```python
# from pymmcore_plus import CMMCorePlus
import pymmcore


# core = CMMCorePlus()
core = pymmcore.CMMCore()

mm_path = .... #"/usr/local/lib/micro-manager"
core.setDeviceAdapterSearchPaths([mm_path])

cb = pymmcore.MMEventCallback()
core.registerCallback(cb)

core.loadSystemConfiguration('demo_config.cfg')

# switch twice to make sure we get at least one switch
core.setConfig('Channel', "DAPI")
core.setConfig('Channel', "FITC")

# you will not see any printouts relating to config changes
```

A use case for the callback always firing is in napari-micromanager where the core state may also be updated by a different python process and it's important for the napari gui to stay up to date.

### Proposed Solution(s)

Either of or some combination of:

1. Remove the check of  `< 1` to allow for notifying when the group only has one property.
2. emit this signal in the `SetConfig` method and then silence/remove the subsequent emission from `SetProperty`
3. Add a new signal `onConfigSet` which fires 

For my money the best option is combine `1` and  `3`. The `onSet` is different from `onChanged` in that the former would only be sent by `SetConfig` and the latter can be triggered by any property change so it makes sense to have both. As for option `1`: it seems that this check was added to support micromanager, but is there any reason that that logic couldn't live in micromanager?

marktsuchida:
Related: https://forum.image.sc/t/micromanager-events-core-events-not-coming-through/53014/6

I think adding a new event with a well-defined behavior is doable. Changing the behavior of the existing `onConfigGroupChanged` is tricky -- even though it doesn't fully make sense, there is complicated GUI code that depends on that behavior.

ianhi:
[...]

PR #26, opened 2021-07-23
create an onConfigSet callback

ianhi:
This will fire when `setConfig` is called successfully. It
is different from `onConfigGroupChanged` because the latter
is called after device properties are set that affect a config
group and has constraints such as not being activated if a config
group only contains one property.

Closes: #25 

Unfortunately I wasn't able to test this as I only know how to test using pymmcore and I was struggling to get it build correctly. But I think this should do the right thing, though perhaps it should be named `onSetConfig`?

### DIFF
--- a/MMCore/MMCore.cpp
+++ b/MMCore/MMCore.cpp
@@ -5062,6 +5062,11 @@ void CMMCore::setConfig(const char* groupName, const char* configName) throw (CM
       throw;
    }
 
+   if (externalCallback_) 
+   {
+      externalCallback_->onConfigSet(groupName, configName);
+   }
+
    LOG_DEBUG(coreLogger_) << "Config group " << groupName <<
       ": did apply preset " << configName;
 }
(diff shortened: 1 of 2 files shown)


marktsuchida:
Appreciate the patch, but I'm not sure this is the right replacement for the deficiency of `onConfigGroupChanged`.

I think it would make more sense to have a callback that notifies when a config group switch is detected due to _any_ cause (either call to MMCore to directly switch the config group or to set a property, _or_ an asynchronous (though unfortunately sometimes synchronous) callback from a device adapter.

(Such a callback could be named `onConfigChanged` to be more consistent with the confusing but largely consistent use of "Config" to mean config preset in the MMCore API.)

If what you need for your use case is actually just a callback that notifies only upon a successful `setConfig` and nothing else, then I would say it is not widely applicable enough to be added to MMCore; you can easily achieve the same result in Python by wrapping the method call.

ianhi:
> If what you need for your use case is actually just a callback that notifies only upon a successful setConfig and nothing else, then I would say it is not widely applicable enough to be added to MMCore; you can easily achieve the same result in Python by wrapping the method call.

Gotcha! I think long term the solutions discussed in #25 will be best, but for now I'm happy to try include something like this in `pymmcore-plus`

PR #57, opened 2021-08-18
Shutter now works on MLDs in a CFLEX

Lukas-Cobolt:


marktsuchida:
Thanks. It looks like some of your temporary scripts got included in the `.vcxproj` file (pre build event, post build event, and outdir) and .gitignore. Would you mind removing these?

Also, did you really intend to include the entire cxxtest project (>400 files)? I'm not sure this is a good idea. Please let me know if I can help with extracting just the changes to the device adapter code into a new branch (it can be done with `git rebase -i` combined with a force-push).

Lukas-Cobolt:
Dear Mark,

Up until recently when you still used SVN, we were able to have our git repo and your svn repo in the same directory. Now when you switched over to git we can no longer control what parts of our repo to publish respectively keep private. The gitignore is an attempt to reduce the amount of files not intended to be published. The reason why I committed it to the repo is to avoid future users (including me) from accidentally committing the now ignored files, causing you further problems.

We use cxxtest to verify our adapter as it is a bit complex and will become more so (it is intended to eventually support all our lasers). We need to have this in our repo, and now as our repo is a fork of your repo, I am not sure how to exclude the test part from the pull request. I have searched for but not found any good solution for this. Do you have any idea?

Regarding the custom scripts, we could create another project file, but it would then be committed too. Is that OK?

Med vänlig hälsning / Kind regards,
Lukas Kalinski

***@***.***
www.coboltlasers.<http://www.coboltlasers./>com
Cobolt AB, a part of HÜBNER Photonics

[Bildresultat för lnkedIn logo]<https://www.linkedin.com/company/80281/>Follow us on LinkedIn!

marktsuchida:
Hi Lukas,

Thank you for the background. I hope the following suggestions will help to come up with something that works for both of us without too much inconvenience.

About the changes to the root `.gitignore` file,

- Visual Studio projects have `X.vcxproj`, `X.vcxproj.filters`, and `X.vcxproj.user`. Of these, the `X.vcxproj.user` should _not_ be committed, but our `.gitignore` has a general rule that excludes all `*.user` files. The `X.vcxproj.filters` is actually part of the project definition and _should_ be committed, so this should not be listed in `.gitignore`.

- If you want to have the file `DeviceAdapters/CoboltOfficial/.gitignore` for your local use, but not commit it, then you can add the line `/.gitignore` to that file itself, so that it ignores itself. The slash makes it specific to the single `.gitignore` file; without the slash it would also ignore all (potential) `.gitignore` files in subdirectories of `DeviceAdapters/CoboltOfficial`.

- But it might be simpler and less error-prone to include `DeviceAdapters/CoboltOfficial/.gitignore` in our official repo. This is what I recommend unless the list of ignored files is itself secret.

So it should be possible to avoid including device-specific entries in the root `.gitignore` file.


About the OutDir and build event scripts in `CoboltOfficial.vcxproj`,

- These must be removed one way or another, because they will cause errors in our automated builds (as well as when anybody else tries to build).

- Not knowing your workflow I can't give full recommendations, but assuming this is for convenience while debugging, you might consider writing a 2-line batch file (or Bash script) that does the copy and launches Micro-Manager (using the `start` command). That way, the process of building, copying, and starting MM is still 2 steps insead of 3.


About the test suite,

- Our general policy is to avoid including third-party projects and binaries in the micro-manager and mmCoreAndDevices repositories, because it causes the repository to get very large and slows down cloing and pulling for everybody. Unlike with SVN, Git downloads and processes the full history, so if these files get included in the main branch once, the effect can never be undone. There are cases where this rule is currently violated, but we want to minimize additional ones.

- Would it cause any problem if you simply added `test/cxxtest` to `DeviceAdapters/CoboltOfficial/.gitignore` and manually place CxxTest in the correct location when needed?

- Alternatively, you could place CxxTest at a standard (to you) location on the computer, and refer to it using absolute paths in the `CoboltOfficialTest` project. You can include the CxxTest version number in the path if you want to ensure that the intended version of CxxTest is always used.

- It is fine to add the `CoboltOfficialTest` project and your test suite and script in our repo. It will not be run by our build scripts, because it doesn't belong to a solution. Since it is not run, it is not a problem if the project refers to absolute paths that only you have on your computer.

PR #85, opened 2021-11-25
First implementation of XCite XT900 NOVEM

S3LSteffen:
As this device takes upto 5 s for closing the serial port, I had to increase timeout for joining the close thread in SerialPort::Shutdown()


nicost:
@marktsuchida, I can not totally oversee the consequences of waiting up to 10 seconds to close the serial port.  Is it OK to merge?

Also, @S3LSteffen, do you have any idea why this can take so long?  I am just curious;)

nanthony21:
is there any possibility of extending the existing Excelitas device adapter libraries to support the new devices rather than creating another semi-diplicate library.

This issue comes to mind: https://github.com/micro-manager/micro-manager/issues/1118

S3LSteffen:
> @marktsuchida, I can not totally oversee the consequences of waiting up to 10 seconds to close the serial port. Is it OK to merge?
> 
> Also, @S3LSteffen, do you have any idea why this can take so long? I am just curious;). B.t.w., code looks great to me!

The 10s are just a timeout for the thread join function. If the close thread is finished earlier, then the join function returns also earlier. So 10s are only for the worst case.  
I did not  investigate why closing this device takes such a long time. I just observed that, also using other serial programs controlling this device.

S3LSteffen:
> is there any possibility of extending the existing Excelitas device adapter libraries to support the new devices rather than creating another semi-diplicate library.
> 
> This issue comes to mind: [micro-manager/micro-manager#1118](https://github.com/micro-manager/micro-manager/issues/1118)

I agree, these duplicates are not good. My problem is that I only have the XT900 and not the other lamps available for testing. So I am curious in changing anything in the existing devices.
Nevertheless, I will compare XT900 and XT600 code and look if there is a way to combine them with low risk...

S3LSteffen:
> > is there any possibility of extending the existing Excelitas device adapter libraries to support the new devices rather than creating another semi-diplicate library.
> > This issue comes to mind: [micro-manager/micro-manager#1118](https://github.com/micro-manager/micro-manager/issues/1118)
> 
> I agree, these duplicates are not good. My problem is that I only have the XT900 and not the other lamps available for testing. So I am curious in changing anything in the existing devices. Nevertheless, I will compare XT900 and XT600 code and look if there is a way to combine them with low risk...

OK, from my code it should be possible to integrate XT900 into XT600.  I will try to do this...

nanthony21:
> OK, from my code it should be possible to integrate XT900 into XT600. I will try to do this...

Thank you, @S3LSteffen. I have an XCite Xylis which uses the XT600 device adapter, so I can use that to test any changes you need to make.

S3LSteffen:
@nanthony21 : I merged XT900 now into XCIteXT600 and removed the separate XCiteXT900.dll
Naming is not really nice... And the controllers would better be hub devices...
...but it is working now. 
Would be nice if you could test with the Xylis.

marktsuchida:
I see no problem with the increased timeout in SerialManager.
Looks okay to merge once @nanthony21 has tested.

marktsuchida:
Reference: https://github.com/micro-manager/micro-manager/issues/1254

nanthony21:
Tested on the Xylis and everything is working as expected. I think we are good to merge, thanks for the contribution!

PR #86, opened 2021-11-30
Add instructions + files for building on linux

ianhi:
This PR adds files to this repo that allow it be built without also downloading the micromanager repo. This is nice for linux users and people using pymmcore (and maybe eventually for pymmcore CI on linux).

I don't really know much about building cpp projects so I mostly just copied the micromanager files and messed with them until I go this working, so I probably did several things non-optimally.

That said if I follow the instructions I end up with a working installation that I can use with `pymmcore`.

Attn @oeway maybe interesting to you as I think this could have implications for https://github.com/micro-manager/pymmcore/issues/19

### DIFF
--- a/building-cpp.md
+++ b/building-cpp.md
@@ -1,0 +1,75 @@
+# Building On Linux
+
+## 1. Install dependencies:
+
+- bzip2
+- bc
+- build-essential
+- cmake
+- curl
+- g++
+- gfortran
+- libboost-dev
+- libboost-thread-dev
+- libtool
+- autoconf
+- automake
+- pkg-config
+- software-properties-common
+- unzip
+- wget
+
+On Debian based OSes you can install these with `apt`
+```bash
+sudo apt update && sudo apt install -y \
+        bzip2 \
+        bc \
+        build-essential \
+        cmake \
+        curl \
+        g++ \
+        gfortran \
+        libboost-dev \
+        libboost-thread-dev \
+        libtool \
+        autoconf \
+        automake \
+        pkg-config \
+        software-properties-common \
+        unzip \
+        wget \
+```
+
+## 2. Build and Install
+
+```bash
+./autogen.sh
+./configure
+make
+sudo make install
+```
+
+The `sudo` is only necessary if you are installing to the default `/usr/local/micro-manager`
+
+## 3. Download MMConfig_Demo.cfg
+Many downstream libraries will expect the demo config file to be located along with the installed files. So here we download it directly from the micro-manager repo:
+
+```bash
+sudo curl https://github.com/micro-manager/micro-manager/blob/master/bindist/any-platform/MMConfig_demo.cfg --output /usr/local/micro-manager/MMConfig_demo.cfg
+```
+
+## 4. Test install
+Now downstream libraries e.g. [pymmcore-plus](https://github.com/tlambert03/pymmcore-plus) should work.
+
+
+```bash
+pip install pymmcore-plus
+```
+
+```python
+from pymmcore_plus import CMMCorePlus
+
+mmcore = CMMCorePlus()
+mmcore.loadSystemConfiguration("demo")
+print(mmcore.getLoadedDevices())
+```
--- a/building-cpp.md
+++ b/building-cpp.md
@@ -55,7 +55,7 @@ The `sudo` is only necessary if you are installing to the default `/usr/local/mi
 Many downstream libraries will expect the demo config file to be located along with the installed files. So here we download it directly from the micro-manager repo:
 
 ```bash
-sudo curl https://github.com/micro-manager/micro-manager/blob/master/bindist/any-platform/MMConfig_demo.cfg --output /usr/local/micro-manager/MMConfig_demo.cfg
+sudo curl https://raw.githubusercontent.com/micro-manager/micro-manager/master/bindist/any-platform/MMConfig_demo.cfg --output /usr/local/micro-manager/MMConfig_demo.cfg
 ```
 
 ## 4. Test install
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,7 @@ test-reports/
 *.pb.cc
 *.pb.h
 bin
+DeviceAdapters/depcomp
+DeviceAdapters/install-sh
+depcomp
+install-sh
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,0 +1,28 @@
+AUTOMAKE_OPTIONS = foreign
+ACLOCAL_AMFLAGS = -I m4
+
+
+# if BUILD_MMCORE
+# MMCORE_DIR = MMCore
+# endif
+
+if BUILD_SECRETDEVICEADAPTERS
+SECRETDEVICEADAPTERS = SecretDeviceAdapters
+endif
+
+
+# CLEANFILES = $(launcher_SCRIPTS)
+
+
+# TODO: Building of DeviceAdapters could be made optional.
+SUBDIRS = \
+	MMDevice \
+	DeviceAdapters \
+	$(SECRETDEVICEADAPTERS)
+
+dox:
+	-rm -rf doxygen/out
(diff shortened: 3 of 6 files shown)


nanthony21:
Thanks @ianhi. I'm not very familiar with the non-Windows side of Micro-Manager, I would just want to make sure that adding these files won't cause problems when building from the top-level Micro-Manager repository, have you checked that?

nanthony21:
I'm also unsure about the changes to .gitignore. Is that needed for this PR or is it specific to your particular situation?

ianhi:
> I'm also unsure about the changes to .gitignore. Is that needed for this PR or is it specific to your particular situation?

Those were all files autogenerated by the build so I added them. It also looks as though they are in the micro-manager repo: https://github.com/micro-manager/micro-manager/blob/e0ba2074bcb95fbf4490815240fe87591f3e3031/.gitignore#L39-L40

> I would just want to make sure that adding these files won't cause problems when building from the top-level Micro-Manager repository, have you checked that?

I haven't checked yet but definitely should. Though notably I am **not** the best person to check that because I've never been able to get it to build correctly with java I always used `--without-java`. That's actually part of the motivation for this PR.

nicost:
I believe this should not interfere with the normal build, but better to check before merging.  Regretfully, it will be a while before I'll have time and access to my Mac.  @marktsuchida is this easy for you to check?

building-cpp.md
```diff
@@ -0,0 +1,75 @@
...
+sudo make install
+```
+
+The `sudo` is only necessary if you are installing to the default `/usr/local/micro-manager`
+
+## 3. Download MMConfig_Demo.cfg
+Many downstream libraries will expect the demo config file to be located along with the installed files. So here we download it directly from the micro-manager repo:
+
+```bash
+sudo curl https://github.com/micro-manager/micro-manager/blob/master/bindist/any-platform/MMConfig_demo.cfg --output /usr/local/micro-manager/MMConfig_demo.cfg
```
ianhi, line 58:
```suggestion
sudo curl https://raw.githubusercontent.com/micro-manager/micro-manager/master/bindist/any-platform/MMConfig_demo.cfg --output /usr/local/micro-manager/MMConfig_demo.cfg
```

marktsuchida:
@nicost I should be able to test this for the Mac build within the next few days.

If it can be (empirically) confirmed that these changes have no effect on the full (micro-manager.git + submodule) build, this seems like a useful transitory solution. I'd prefer if it is made more clear (to somebody trying to build for the first time) that these files are an alternative to the main (micro-manager.git) configure/make and that they only build the devices, not MMCore (contrary to the name of the repository) (did I read that correctly?). Perhaps by printing an explanation from `autogen.sh`.

Minor: The check for SWIG can be removed. Since this method is unlikely to ever be used to build the closed-source device adapters, all references to SecretDeviceAdapters (except for the `mkdir` in `autogen.sh`, which might be needed) should be okay to remove.

ianhi:
>  I'd prefer if it is made more clear (to somebody trying to build for the first time) that these files are an alternative to the main (micro-manager.git) configure/make and that they only build the devices,

Makes sense will do.

>  and that they only build the devices, not MMCore (contrary to the name of the repository) (did I read that correctly?). 

ooh - it was my intention to build those as well, as can be seen in the configure file, but I maybe left them out of the makefile? I just assumed that they had built because pymmcore work, but is that because pymmcore includes mmcore in itself?


> this seems like a useful transitory solution

Out of curiosity Transitionary to what?


> Since this method is unlikely to ever be used to build the closed-source device adapters, all references to SecretDeviceAdapters

See also https://github.com/micro-manager/mmCoreAndDevices/issues/87 there currently is no way to get the secret device adapters on linux. It would be great to have a method to be able to get that.

marktsuchida:
> ooh - it was my intention to build those as well, as can be seen in the configure file, but I maybe left them out of the makefile? I just assumed that they had built because pymmcore work, but is that because pymmcore includes mmcore in itself?

Actually, seeing that this is intended for pymmcore users, it makes sense to leave out MMCore, because the only "official" build of MMCore is as MMCoreJ. MMCore can also be built as a C++ library, but that is not really supported by the current build system, for now. Pymmcore currently builds MMCore from the source files using setuptools and does not use the autotools scripts or Visual Studio projects.

> Out of curiosity Transitionary to what?

We (@nanthony21 and I) are starting work on a Meson-based build for the C++ components. In my vision, each device adapter will have its own Git repository and CI. The hope (among other things) is that this will vastly simplify common tasks for the majority of developers, who only work on a single module at a time. There will be a separate mechanism to aggregate all the adapters.

> See also #87 there currently is no way to get the secret device adapters on linux. It would be great to have a method to be able to get that.

That would be nice, but I think the problem is orthogonal to building conda packages (or at least I would expect it to be more work to build correct conda packages than to merely build Linux binaries for distribution -- but I could be wrong). For some adapters (serial-based ones), it is feasible to build binaries that will work with (almost) any distro. For others, only a specific Linux distro is supported by the vendor's driver. The number of important closed-source devices that can work on Linux is small, though.

tractatus:
@marktsuchida any updates on the Meson build? I've been starting to clean up the original autotools to see if a minimalistic and clean CMake build can be done for Linux and macOS and only then taking on Windows since that usually requires some refactoring of the CMake to fit Windows.

But if Meson build is the future of the MMCore C/C++ then it probably is better for me to invest in that.

OT this sounds awesome:
`In my vision, each device adapter will have its own Git repository and CI. `

marktsuchida:
@tractatus We decided to work on updating the Windows compiler first (`vs2019` branch), but that's still the plan!

tractatus:
@marktsuchida got it! Ok I think I will then spend some time and port all my Cmake builds of what I'm working on to Meson to be more in line with how MMCore develops. 

@ianhi how does your situation currently looks like with *unix build of MMCore? My experience was pretty similar to yours fiddling around with the automake and then making the shared library work together with a very lightweight CMake build of some C/C++ I have that uses Imgui as GUI and OpenCV for image processing of the image buffer.
https://github.com/tractatus/MMCore_OpenCV_Imgui

ianhi:
> @nicost I should be able to test this for the Mac build within the next few days.

@marktsuchida were you ever able to check how/if this affects the mac build?

> @ianhi how does your situation currently looks like with *nix build of MMCore? My experience was pretty similar to yours fiddling around with the automake and then making the shared library work together with a very lightweight CMake build of some C/C++ I have that uses Imgui as GUI and OpenCV for image processing of the image buffer.

I'm a huge noob w.r.t C++ unfortunately so this is about the extent of what I've done. For me getting the device drivers was sufficient. I'm pretty invested in pymmcore-plus/napari-micromanager at the moment. Most of my use cases aren't data rate limited but more complexity of control limited (automatically updating laser aiming) so python is a nice place to be.

> I have that uses Imgui as GUI and OpenCV for image processing of the image buffer.

That's very cool!

marktsuchida:
Please also see https://github.com/micro-manager/micro-manager/issues/1392#issuecomment-1033140970 for future build system plans.

tractatus:
@marktsuchida thanks for the update. So I'm a little bit confused this means that `mmCoreAndDevices` will transition over to Mason as build system with support for Linux or this will not happen?

marktsuchida:
@tractatus Meson will be able to configure and build for all supported platforms, using a single set of scripts (in contrast to using `*.vcxproj` on Windows (with a configuration mechanism lacking) and `configure.ac` + `Makefile.am` on Linux/macOS). So yes, the individual modules within `mmCoreAndDevices` will each be built using Meson, including for Linux. (From the user's perspective, `./configure` + `make` will be replaced by `meson` and `ninja`.)

ianhi:
sounds good! may open a different PR with instructions on how to make the configure route work so it's easier for the next person (if this imagined person comes along before the new build system does)

marktsuchida:
@ianhi That would be cool, because I think the new build system will take some time to materialize. And there are always some system-specific fine points when running `./configure`.

PR #94, opened 2021-12-04
Remove projects that aren't included in the solution file.

nanthony21:
These projects are not included in the solution file so I see no reason to keep them around.


nicost:
ArduinoFuncGen - written by Mark in 2014, so he should decide
MicroFPGA - by Joran Deschamps in 2020. We possibly forgot to include this in the build. I'll ask him directly
ServiceZDrive - by Henry Pinkard.  Let's ask him.
iPadSLM - by Arthur and Mark.  What do you think Mark?

henrypinkard:
iPadSLM was also mine. A quick and dirty copy of the GenericSLM which is maybe why it has a misleading commit history. We can definitely remove it since neither I nor anyone else is using it

nanthony21:
Ok, the removal of servozdrive was confirmed separately so it sounds like we're just waiting to hear back about:
- [ ] arduinoduefuncgen
- [ ]  microfpga

marktsuchida:
You can remove ArduinoDueFuncGen. It worked and is fairly complete for its original purpose, but assumes a very specific DAC to be used with the Arduino, so not very likely that others will use. Good that it's in the history but no need to keep it in the working tree.

marktsuchida:
Once this is merged to `vs2019`, let's cherry-pick this commit on `main`. (That way, we prevent conflicts arising from somebody modifying the removed projects on `main`.)

nanthony21:
The author of MicroFPGA plans to submit some updates to it soon. For now let's just leave it as is. I think this is ready to merge now. We can squash merge to get rid of the history of having removed microFPGA

PR #97, opened 2021-12-07
Remove deprecated minimal rebuild option.

nanthony21:
The usage of the "C++ -> Code Generation -> Enable Minimal Rebuild" option in some projects triggers a deprecation warning when using v142 platform toolset. See here: https://docs.microsoft.com/en-us/cpp/build/reference/gm-enable-minimal-rebuild?view=msvc-170

This PR makes the projects all inherit from the default to not use this option.


marktsuchida:
Great. Can you make a corresponding PR for SecretDeviceAdapters, too?

PR #101, opened 2021-12-09
Remove updatePrivateMain workflow and replace with script

marktsuchida:
This adds a file containing the corresponding commit of SecretDeviceAdapters so that automated builds can check out the correct code. This is an alternative to using a submodule; a submodule would be annoying for all who do not have access to SecretDeviceAdapters.

The file `secret-device-adapters-commit` is to be kept up to date by a workflow that runs when there is a push to SecretDeviceAdapters.


marktsuchida:
@nanthony21 `SecretDeviceAdapters` is already in the `.gitignore.`.

I don't know what `git submodule update --recursive` says when the credentials are missing, but at the very least it will show an error, which is not desirable. It may ask for a password, which is even more annoying. I think we should make it a rule that submodules of public repos are all public. In the future I'm hoping we may be able to address this kind of situation (e.g. proprietary dependencies of public device adapters) using Meson wrap files instead of submodules.

PR #104, opened 2021-12-12
PointGrey for vs2019

nanthony21:
The PointGrey adapter is fixed by replacing the no longer existent `_ULongLong` identifier with `unsigned long long`.

Remaining question: Why cast to `unsigned long long` in the first place? Why not just pass the `unsigned int` variable directly.


nicost:
I probably wrote that code, but have no recollection.  Probably easiest to leave as you propose.  I am happy to test some time in the future, as I'll likely continue to have access to one of these cameras.

PR #107, opened 2021-12-14
vs2019: Remove linker 'VERSION' flag from all projects

nanthony21:
Two of the XCite project files had a "Version" option in the linker configuration which results in a syntax error in VS2019.

The version hasn't been updated in more than a decade and I doubt anyone knows it is there any way. This PR removes the version and enables building these device adapters on VS2019.


marktsuchida (approved):
Not sure if it was introduced intentionally or not, but it is clear we don't need it.
I found another project (MP285) so added a commit.

PR #110, opened 2021-12-15
update TIScam SDK

### LINKED ISSUE #106: Fix TIScam for new msvc version
The TIScam device adapter currently fails to build  with msvc v142 due to the following lines in the SDK that it includes from "3rdParty":
```C++
#elif _MSC_VER > 1800
  #error This compiler was not tested with this library.
#else
  #error Wrong Compiler. This library does only run with Visual C++ 7.1, 8.0, 9.0 and 10.0.
  // other maybe newer compiler ...
#endif
```

If the `#error` is commented out then everything builds fine. I can commit this change to the SVN repo but wanted to use this issue as a place to review the change before commiting.

nanthony21:
This PR updates the "TheImagingSource" SDK used by the TIScam device adapter to be compatible with newer compilers.

This PR will require rev 661 of the `3rdParty` SVN repository.

This has not been tested on hardware.

closes #106


marktsuchida:
Looks good! We'll have to depend on users to test on hardware, but if you haven't yet, it might be good to check that the device adapter loads without errors when opening the Hardware Configuration Wizard.

PR #115, opened 2021-12-20
Debug build fixes

nanthony21:
This PR fixes a few projects that build find in "Release" mode but not in "Debug". There are still a few projects remaining, the "Mightex" and "Mightex_BLS" are the main ones that I'm not sure what the issue is.

### DIFF
--- a/DeviceAdapters/NewportCONEX/NewportCONEX.vcxproj
+++ b/DeviceAdapters/NewportCONEX/NewportCONEX.vcxproj
@@ -44,7 +44,7 @@
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup>
     <_ProjectFileVersion>10.0.40219.1</_ProjectFileVersion>
-    <LinkIncremental Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">false</LinkIncremental>
+    <LinkIncremental Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">true</LinkIncremental>
     <LinkIncremental Condition="'$(Configuration)|$(Platform)'=='Release|x64'">false</LinkIncremental>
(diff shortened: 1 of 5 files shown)


marktsuchida:
Some projects do not build correctly with the Debug runtime (whether or not there are any build errors). Has this been taken into account?

nanthony21:
I have only changed the runtime for projects which were already failing to build with errors like this: `LNK2001	unresolved external symbol __imp__CrtDbgReport	NewportCONEX	C:\Users\nicke\Documents\Bitbucket\micro-manager\mmCoreAndDevices\DeviceAdapters\NewportCONEX\MMDevice-SharedRuntime.lib(DeviceUtils.obj)	1	
`. After switching the runtime these project compiled without any errors, whether they actually run hasn't been tested.

marktsuchida:
Ok, in that case this should be fine to merge.

The Debug build has been poorly maintained (because it is not built in CI and never shipped), and nobody has ever done the work to make every project work. Sometimes people fix it when they need it for a particular project. Sometimes people randomly break it.

To do it correctly, all projects should use Debug builds of their dependencies where available (some vendors supply them). Where not available and the dependency has a C API, the Release build of the dependency can (usually) be used with the Debug build. When the dependency has a C++ interface, the device adapter should use the matching (non-Debug) runtime even in the Debug build. In the latter case, the project will build without optimizations in Debug mode, allowing debugging, even though it won't have the special features of the Debug runtime (such as detecting memory errors).

Probably not worth spending too much effort until we move to meson, though. The main problem in VS (unless we develop our own check scripts) is that we don't have a mechanism to detect when people forget to make the correct changes to all of the (Platform-)Configuration combinations -- which happens quite frequently (and understandably).

PR #120, opened 2022-01-05
Fix build of Mightex and Mightex_BLS for VS2019

nanthony21:
When building in debug mode these two device adapters would fail, apparently due to conflicting declarations found in the Windows 10 SDK and in the version of WinDDK included in the 3rdParty repository.

This PR fixes the VS2019 build by removing the WinDDK from the include path of the projects.

I'm not 100% sure but I think this means that WinDDK (1.3GB) could be removed from the 3rdParty repository.


marktsuchida:
Looks good.
As we already discussed, nothing should be removed from 3rdparty until it is retired entirely.

PR #124, opened 2022-01-18
Update TSI device adapter load files to work with SDK 2.0.1 DLLs

jfriesen-thorlabs:
SDK has new functions and a reworked load file for DLLs. Changes are required for TSI camera users with SDK v2.0.1.


marktsuchida:
I will merge this after updating `3rdparty` and bumping `3rdparty-revision` in the micro-manager repo.

PR #132, opened 2022-01-27
Fix PicardStage timing out

f-marelli:
Due the lack of default value for homing in the CPiStage, it is randomly considered as indefinitely busy when loading a configuration.
Adding the default value (false) should prevent random device timeouts at launch.


marktsuchida:
Thank you!

PR #139, opened 2022-02-04
IDS_uEye: updated ROI position calculation

mueller-physics:
IDS uEye cameras limit ROI positions and sizes to certain increments, e.g. multiples of 8 in width and 2 in height for UI-3060 cameras. There is also a minimal size for ROIs.

The current device adapter does not check for supported ROI position, and (in our testing) sending unsupported positions is silently ignored (i.e. corrected to a nearby supported position) by the camera. **This yields incorrect **metadata**, which we notice when using e.g. pixel-based image corrections.

The pull request reworks the SetROI(...) call to send only supported ROI sizes and positions to the camera.


marktsuchida:
Thank you for the fix!

PR #142, opened 2022-02-09
IDS_uEye additional features

mueller-physics:
Porting some additions we made to MM 1.4 IDS uEye device adapter to MM 2.0 code base.
https://github.com/biophotonics-bielefeld/ids-device-adapter

This PR is three commits, adding:

* User-settable camera ID, allows for running multiple IDS cameras in MM
* Option to control the 'flash output' (TTL-style exposure state output) of the cameras
* In-camera timestamp (and IO port state) gets added to metadata in sequence acquisitions

All of these *should* be backwards compatible with current configurations, as default values (camera id 1, flash off) match those used by the current adapter.


marktsuchida:
Thanks!

PR #145, opened 2022-02-10
Add new DACXYStage device to ASITiger

bls337:
This adds functionality for the DAC_XY_4CH firmware build.


bls337:
@nicost @marktsuchida Do I need to add ASIDacXYStage.h and ASIDacXYStage.cpp to ASITiger.vcxproj and ASITiger.vcxproj.filters for this to work? What about Makefile.am, it looks like I need it there too, is that part of the build process for Linux?

marktsuchida:
@bls337 Yes, please add the new files to both the VS project and the `Makefile.am`.

bls337:
@marktsuchida Alright, I fixed those files, it should be good now.

Issue #37, opened 2021-08-02
MMCoreJ doesn't build with Swig 4.0


marktsuchida:
This is a continuation of micro-manager/micro-manager#838.

Quoting from the [Swig Changelog](http://www.swig.org/Release/CHANGES), in 4.0.0 there is the following entry:
```
2019-02-28: wsfulton
            [Java] std::vector improvements for types that do not have a default constructor.

            The std::vector wrappers have been changed to work by default for elements that are
            not default insertable, i.e. have no default constructor. This has been achieved by
            not wrapping:

              vector(size_type n);

            Previously the above had to be ignored via %ignore.

            If the above constructor is still required it can be added back in again via %extend:

              %extend std::vector {
                vector(size_type count) { return new std::vector< T >(count); }
              }

            Alternatively, the following wrapped constructor could be used as it provides near-enough
            equivalent functionality:

              vector(jint count, const value_type& value);

            *** POTENTIAL INCOMPATIBILITY ***
```
This suggests that it might be possible to have code that works across Swig versions.

JiangXL:
May I ask how it is going with Swig 4.0?

marktsuchida:
Let me try to at least make a to-do list for this.

Below are the relevant changes (swig <4 to swig 4) that we need to deal with (generated using [japicmp](https://github.com/siom79/japicmp) and filtered).

They result from two changes to SWIG's Java support (from the changelog linked above):
- 2017-05-26 "Extend from java.util.AbstractList<> and implement java.util.RandomAccess for std::vector wrappers"
- 2019-02-28 "std::vector improvements for types that do not have a default constructor"

Of the API changes, we can accept the ones labeled NEW INTERFACE, MODIFIED SUPERCLASS (from `Object`), NEW CONSTRUCTOR, and NEW METHOD. We can also accept the single NEW CLASS (`StrMap$Iterator`). (But it does mean we should switch to _requiring_ SWIG 4 once we enable it, or else the API will become incompatible between copies of `MMCoreJ.jar` built by different SWIG versions.)

Among the backward-incompatible changes, there is not much we can do about the MODIFIED METHODs, where the return value of `get()` got changed to the boxed types and the return value of `size()` got changed from `long` to `int`. These can break existing user code, but hopefully not that frequently; the only failure mode is a compile error, so people will easily see how their code needs to change.

That leaves us with the REMOVED CONSTRUCTORs and REMOVED METHODs. More about these in my next comment.

```
[...]

PR #146, opened 2022-02-11
Fix type cast errors that occur when compiling with Swig 4.0.

fabio-echegaray:
Avoids the type cast errors that are thrown when casting number types (e.g. Long to Integer) by using HashMap java class.

Compilation goes well under Linux; haven't tested it under Windows or MacOS. This change might potentially need some further testing. I am happy to keep updating the PR if required.

### DIFF
--- a/MMCoreJ_wrap/MMCoreJ.i
+++ b/MMCoreJ_wrap/MMCoreJ.i
@@ -633,11 +633,15 @@
    import mmcorej.org.json.JSONObject;
    import java.awt.geom.Point2D;
    import java.awt.Rectangle;
+   import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


marktsuchida:
Thanks for looking at this issue!

Does this keep backward compatibility of the Java API of MMCoreJ? Please see #37.
(One thing we are sorely lacking is unit tests to ensure that the API doesn't change.)

I would expect macOS to behave like Linux. Windows has a 32-bit C++ `long`, so will need to be tested.

fabio-echegaray:
Aha! I'll work on unit tests for the integration of MMCoreJ to the Java API then.

PR #157, opened 2022-02-21
Add links to api docs to readme

ianhi:
I always go here first to find these, then remember I have to go pymmcore in order to find a link. Nice to also have them here.

I added the second link because the main page is currently empty of content (ill open an issue about that) and CMMCore is probably the thing most people will be looking for.


marktsuchida:
Thanks (also for the other good issues you just created)!

PR #167, opened 2022-03-01
NIDAQ device adapter

nicost:
Based on NI Multi Analog adapter.  Adds digital output ports.  Includes blanking.  tested with PCIe-6251, USB-6008, and PCIe-6353 (emulator only, port1 did not work?).  User documentation to follow.


marktsuchida:
I'll hopefully look at this tomorrow.

marktsuchida:
Closes #51.

nicost:
All great points Mark!

1. Naming:  Note that the name becomes `NIDAQDO-Dev1/port0-/line0` with the second dash inserted by Micro-Manager to separate device and property name.  I think that I prefer `NIDAQDO-Dev1/port0-line0`.  Still close enough to the NIDAQ nomenclature, and a tiny bit less clutter.  

2. I took the second option, namely let Initialize reduce the number of sliders.  Reason is that I do not like to run the code to test if the port can do change-detection in the constructor.

nicost:
Think that I got most of the error reporting issues.  The one I can test easily (port 1 on emulated PCIe-6353 does not work for whatever reason), now shows a nice error message (which previously was a pain to get to;).

marktsuchida:
> * Naming:  Note that the name becomes `NIDAQDO-Dev1/port0-/line0` with the second dash inserted by Micro-Manager to separate device and property name.  I think that I prefer `NIDAQDO-Dev1/port0-line0`.  Still close enough to the NIDAQ nomenclature, and a tiny bit less clutter.
> * I took the second option, namely let Initialize reduce the number of sliders.  Reason is that I do not like to run the code to test if the port can do change-detection in the constructor.

Totally agree with both decisions! (I would have gone with `NIDAQDO-Dev1-port0-line0`, but maybe that would be bikeshedding....)

> port 1 on emulated PCIe-6353 does not work for whatever reason

I saw the same thing when simulating a PXIe-6356 (X Series). `port0` and `port2` work but `port1` appeared to be in use. It started working after clicking Reset in NI MAX (I did the reset after loading the device in MM). It's probably some default setting getting in the way, but I did not investigate further.

PR #170, opened 2022-03-02
ASIStage: refactor into classes for each device

bls337:
Improvements to the organization of the ASIStage device adapter.


marktsuchida:
@bls337 Is it okay if I "squash and merge" this (so that the whole PR becomes a single commit in the history)?
I think this will make sense because your individual commits appear to be separated by files they touched rather than the code being moved, and are therefore only the last commit will actually build.

bls337:
Yeah that works for me, thanks.

bls337:
@marktsuchida tagging you so that you see me reply.

marktsuchida:
Thanks.

PR #190, opened 2022-04-28
[WIP] Add callbacks to Prior Stage Adapter

ianhi:
Open question is how to best do the relative callbacks?

### DIFF
--- a/DeviceAdapters/Prior/Prior.cpp
+++ b/DeviceAdapters/Prior/Prior.cpp
@@ -1004,6 +1004,7 @@ int XYStage::SetPositionSteps(long x, long y)
 
    if (answer.substr(0,1).compare("R") == 0)
    {
+      OnXYStagePositionChanged(x*stepSizeXUm_, y*stepSizeYUm_);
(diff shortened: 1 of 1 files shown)


marktsuchida:
Micro-Manager device adapters for motorized stages are supposed to return from the SetPosition functions after just kicking off the movement (and possibly receiving a command acknowledgment). User code then waits for completion of movement by polling `Busy()`. So the very idea of issuing position callbacks from the SetPosition functions is probably flawed, at least for most stages.

In the case of integrated motorized microscope stages (such as those from the Big 4), the serial (or other) API typically has asynchronous messages to indicate end of motion, at which point callbacks can be generated. But this is not possible with many more traditional stage interfaces, where motion completion can only be detected by polling.

I'm not sure it is possible to do this correctly without a thread that polls the device for motion completion. It is questionable, however, if it is a good idea for each of these device adapters to reinvent such a complex mechanism. In fact, the StageControl window performs some polling for this very reason, IIRC.

(Let's use GitHub's "draft" PR feature for WIP PRs.)

PR #193, opened 2022-05-04
Add support for Sutter Lambda 721

dscbiggs:
Add support for Sutter Lambda 721 multi-LED light source as a Shutter device.
Allows single LED mode (same as a filter wheel), and multi-LED mode allowing up to 7 LEDs activated.  Also provides individual power % level.

### DIFF
--- a/DeviceAdapters/SutterLambda/SutterLambda.cpp
+++ b/DeviceAdapters/SutterLambda/SutterLambda.cpp
@@ -80,7 +80,7 @@ const char* g_Keyword_Power7 = "Power LED 7";
 
 const char* g_Keyword_LEDPosition = "LED Selected";
 const char* g_Keyword_LEDMulti = "Multi-LED Mode";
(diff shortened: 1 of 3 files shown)


marktsuchida:
Thanks!

Just one thing: are the `#ifdef`s for `DefineLambda721` needed for something? Unless there is a need for it, I'd like to remove them to keep the build simpler (in particular, the macOS/Linux builds won't get the 721 code without additional configuration now).

dscbiggs:
The define isn't needed.  I was just keeping the code separate incase I broke something or wasn't to be included.
You can remove them, or I can update the pull request.  Thanks for reviewing!

marktsuchida:
@dscbiggs Thanks, would be great if you could do it (also maybe remove commented code where unlikely to be needed), but I can also do it tomorrow.

dscbiggs:
I removed the define and unnecessary comments.  Thanks.

PR #197, opened 2022-05-10
Sutter lambda fix unix build

marktsuchida:
Fix Unix build following #193.

@dscbiggs Could you take a look at my first commit ('Remove redundant/dead code') and make sure it is correct? I'm not sure what the role of the second parameter (`id`) to the `ShutterOn721` constructor is. There were lines setting `id` to `0` and `3`; I removed the one using `3` because it never would have been reached.

(Also, just to check, is it correct to pass `3` as the `id` to the `Wheel` constructor in the case of `g_WheelLambda721Name`?)


marktsuchida:
I went ahead and merged this to fix the build; if fixes are needed for the `id` parameters, please create a new PR.

PR #200, opened 2022-05-15
New Andor Dragonfly device adapter

cmessaouik:
Introduction of the Andor Dragonfly device adapter developped in collaboration with Andor.

The adapter is dependent on the Andor ASD SDK. The SDK library is loaded dynamically so only the SDK header files will be required to build the device adapter. The SDK is expected to be sent to you shortly once Andor has given their final authorisation.

This is a Windows 64b only device adapter. The project will likely not build on Linux. Let me know if it has to be buildable on Linux (however, note that it won't be usable since the dependent SDK is Windows-only).


marktsuchida:
@cmessaouik Sounds good; please email me (and @nicost) when you can provide the SDK.

No need to worry about Linux if the SDK is Windows-only.

Can you also indicate the copyright and license (best to include a `license.txt` file)?

cmessaouik:
Work rebased on main and new license file added.
I will email you the SDK shortly.

DeviceAdapters/Dragonfly/ConfigFileHandler.cpp
```diff
@@ -0,0 +1,137 @@
...
+  : MMDragonfly_( MMDragonfly )
+{
+  FileName_ = "Undefined";
+
+  // Create property with a default path set to My Documents
+  PWSTR vWPathName;
+  if ( SUCCEEDED( SHGetKnownFolderPath( FOLDERID_ProgramData, KF_FLAG_DEFAULT, NULL, &vWPathName ) ) ) // FOLDERID_PublicDocuments for Public documents path - FOLDERID_ProgramData for All users (aka ProgramData) path
+  {
+    wstring vWPathNameString = vWPathName;
+    FileName_.assign( vWPathNameString.begin(), vWPathNameString.end() );
```
marktsuchida, line 29: Here I get the warning

```text
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\xstring(2541,23): warning C4244: 'argument': conversion from 'wchar_t' to 'const _Elem', possible loss of data
        with
        [
            _Elem=char
        ]
```

as a result of passing a pair of wchar_t iterators to std::string::assign (it will not autoconvert but rather take the low byte of each wide character).

I think this is safe if the string is guaranteed to be ASCII only. Is that a good assumption here (e.g. on non-English versions of Windows)?

marktsuchida (commented):
Dependencies are now in place in `3rdparty`.

I saw 1 compiler warning (see below) that I think is worth looking at.

As with the ILE adapter (#201), the `.vcxproj.filters` file is missing; great if you can add it if possible.

Other than that, this is ready to merge.

DeviceAdapters/Dragonfly/ConfigFileHandler.cpp
cmessaouik: Good point. I now use ```wcstombs_s``` to do the conversion from ```wstring``` to ```string```. From what I gathered it may not be bulletproof but there's an easy workaround for the user if the conversion doesn't work properly.

cmessaouik:
@marktsuchida The filter file was added

marktsuchida:
@cmessaouik Thanks!

PR #201, opened 2022-05-15
New Andor ILE device adapter

cmessaouik:
Introduction of the Andor ILE device adapter developped in collaboration with Andor.

This adapter has similar constraints as the Andor Dragonfly adapter.
It depends on the Andor ALC SDK from which only the header files are required to build the adapter. The SDK is expected to be sent to you shortly once Andor has given their final authorisation.
This is also a Windows 64b only device adapter. Let me know if it has to be made buildable for Linux.


marktsuchida:
@cmessaouik Sounds good; please email me (and @nicost) when you can provide the SDK.

No need to worry about Linux if the SDK is Windows-only.

Can you also indicate the copyright and license (best to include a `license.txt` file)?

nicost:
HI @cmessaouik, any updates on availability of the SDKs?  Would be nice to incorporate your work!

cmessaouik:
Sorry @nicost for the delay. I added the license file and rebased on main.
I will send you the SDK files shortly.

DeviceAdapters/IntegratedLaserEngine/ILEWrapper/ILEWrapper.cpp
```diff
@@ -0,0 +1,367 @@
...
+}
+
+void UnloadILEWrapper()
+{
+  MMThreadGuard g( ILEWrapperLock_s );
+  --CILEWrapper::NbInstances_s;
+
+  if ( CILEWrapper::NbInstances_s == 0 )
+  {
+    delete ILEWrapper_s;
```
marktsuchida, line 65: I get the warning

```text
...\DeviceAdapters\IntegratedLaserEngine\ILEWrapper\ILEWrapper.cpp(65,5): warning C5205: delete of an abstract class 'IILEWrapperInterface' that has a non-virtual destructor results in undefined behavior
```

which I think can be fixed by simply adding `virtual ~IILEWrapperInterface() = default;` to `ILEWrapperInterface.h`.

DeviceAdapters/IntegratedLaserEngine/PortsConfiguration.cpp
```diff
@@ -0,0 +1,261 @@
...
+  return vBaseFileName + Device1Name + ", " + Device2Name + ".desc";
+}
+
+bool CPortsConfiguration::GetFusionConfigFolderPath( std::string* Path ) const
+{
+  PWSTR vWPathName;
+  if ( SUCCEEDED( SHGetKnownFolderPath( FOLDERID_PublicDocuments, KF_FLAG_DEFAULT, NULL, &vWPathName ) ) ) // FOLDERID_PublicDocuments for Public documents path - FOLDERID_ProgramData for All users (aka ProgramData) path
+  {
+    std::wstring vWPathNameString = vWPathName;
+    std::string vBaseFileName( vWPathNameString.begin(), vWPathNameString.end() );
```
marktsuchida, line 86: Here I get the warning

```text
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\xstring(2541,23): warning C4244: 'argument': conversion from 'wchar_t' to 'const _Elem', possible loss of data
        with
        [
            _Elem=char
        ]
```

as a result of passing a pair of `wchar_t` iterators to the constructor of `std::string` (it will not autoconvert but rather take the low byte of each wide character).

I think this is safe if the string is guaranteed to be ASCII only. Is that a good assumption here (e.g. on non-English versions of Windows)?

marktsuchida (commented):
Dependencies are now in place in `3rdparty`.

I did not review the code in detail, but saw 2 compiler warnings about which I commented below. I think they are worth looking at.

Other than that, this is ready to merge.

marktsuchida:
Also it looks like the `.vcxproj.filters` file is missing. This does not affect the build but if you could add it the project will show up more correctly in Visual Studio. (In contrast to `.vcxproj.user`, which contains local user settings, the `.vcxproj.filters` file is meant to be tracked by version control.)

DeviceAdapters/IntegratedLaserEngine/ILEWrapper/ILEWrapper.cpp
cmessaouik: Good catch. Fixing this revealed a bug in the destructor of ```ILEWrapper```.

DeviceAdapters/IntegratedLaserEngine/PortsConfiguration.cpp
cmessaouik: I improved this the same way as in the Dragonfly adapter using ```wcstombs_s```.

cmessaouik:
@marktsuchida I modified the project following your various remarks. I took this opportunity to change the 2 latest commits messages to make them start with a capital letter hence the force push.

marktsuchida:
@cmessaouik Perfect, thanks.

Issue #186, opened 2022-04-20
update libusb support


artmeln:
I’m writing a device adapter that uses libusb as a library. After looking at 3rdpartypublic, it looks like libusb versions utilized by MM device adapters in the past are rather outdated. Would it be possible to add libusb-1.0 to 3rdpartypublic? It is the current platform independent version of the library.

marktsuchida:
I think we can make use of libusb 1.0 without necessarily updating the existing device adapters that need libusb 0.1 (aka libusb-compat). If you contribute a device adapter that uses libusb-1.0, we should be able to add a Windows binary to 3rdpartypublic (and a scripted build for macOS).

Let's not add dependencies until they are actually used, however.

Also, it will be good to make sure that talking directly to libusb is the right thing to do, i.e., that you are not reinventing what USBManager does.

artmeln:
libusb (version 1.0.26) has been made available in 3rdpartypublic

PR #202, opened 2022-05-15
Adding a new adapter UniversalMMHubUsb

artmeln:
There are two changes here:
1. My previously contributed UniversalSerialHub adapter has been renamed UniversalMMHubSerial for consistency.
2. A new adapter UniversalMMHubUsb has been added. This adapter relies on the libusb library, so I added a hard-coded path  to the Visual Studio project settings for now (my local /projects/3rdpartypublic/libusb-1.0.26). It seems that adding this dependency should be possible (https://github.com/micro-manager/mmCoreAndDevices/issues/186), and this adapter requires libusb to achieve high-speed bulk USB transfer


artmeln:
I've updated both directories with the license files. Thanks for the suggestions, there is lot of code duplication between the USB and serial versions. I'll look into reorganizing based on the examples you mentioned.

marktsuchida:
Thanks. I'll merge this once I've had a chance to add libusb-1.0. (You can do any reorganization later; I won't block this PR for that.)

marktsuchida:
@artmeln Thanks for your patience. Would be great if you could check the nightly build (20220603 or later) to make sure it works. I used the static library from the official Windows binary package.

artmeln:
I've verified that everything works as expected with the latest build. Thank you for your help.

PR #208, opened 2022-05-27
fix bugs in SpinnakerCamera ROI settings

ieivanov:
This PR fixes bugs during ROI setting of Spinnaker cameras. For example, previously an error would arise if attempting to set the ROI to (0, 0, 100, 100) due to the way increments in OffsetX and OffsetY are handled. This PR addressed this bug and as a result would set ROI position that is slightly different (<OffsetX.GetInc() pixels) compared to the previous version.


marktsuchida:
Thank you!

PR #223, opened 2022-07-25
Dragonfly: Fix filter wheel position description

cmessaouik:
If a filter wheel position string length is bigger or equal to the
buffer it is going to be stored in, it is missing a terminal 0.
This leads to scrambled text stored in the property value preventing
the use of this value.


cmessaouik:
Sorry, I just realised this fix was missing in the original Dragonfly pull request.

PR #232, opened 2022-08-08
Fix Zeiss uninitialized memory issue

wxs:
I was having an issue on an x64 Linux computer
with incorrect absolute XY stage positions being read. After some investigation I tracked it down to it accessing uninitialized memory due to an invalid cast.

ZeissLong is only 4 bytes, but on (most?) 64-bit machines a long is
8 bytes. There were parts of the code that were casting a long
to a ZeissLong leading to undefined values for values that were
being read.

I tried to clean it up in a few positions, although the specific bug that I was facing stemmed from XYStage::GetPositionSteps: 

https://github.com/micro-manager/mmCoreAndDevices/compare/micro-manager:300caa5...wxs:c5a8242#diff-f2bd135a3e8514850d9899bb795dc4dbd273a9bb1ef492871da41077b33cd027R1891

I also changed some typedefs to make the integer sizes explicit, as this could be an issue on different architectures.

Side comment: `ZeissCan29.cpp` and `.h` seem to use Windows line endings, but everything else seems to use Unix line endings. I left the line endings the same to make it easier to read the diffs, but seems like this should be made consistent?

### DIFF
--- a/DeviceAdapters/ZeissCAN29/ZeissCAN29.cpp
+++ b/DeviceAdapters/ZeissCAN29/ZeissCAN29.cpp
@@ -1996,7 +1996,7 @@ int XYStage::OnTrajectoryVelocity(MM::PropertyBase* pProp, MM::ActionType eAct)
    } else if (eAct == MM::AfterSet) {
       double tmp;
       pProp->Get(tmp);
-      ZeissLong velocity = tmp * 1000.0;
+      auto velocity = ZeissLong(tmp * 1000.0);
       int ret = ZeissAxis::SetTrajectoryVelocity(*this, *GetCoreCallback(), g_StageXAxis, velocity);
       if (ret != DEVICE_OK)
          return ret;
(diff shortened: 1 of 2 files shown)


marktsuchida:
Thanks for this.

`long` is 4 bytes on 64-bit Windows, so let me give this a check on Windows before merging.

Regarding newlines, we should fix it at some point, but it will be good to do the whole repo at once (I recently [did it](https://github.com/micro-manager/micro-manager/pull/1469/commits/f4d24e2b1225228509b2291ebc6b1bdfca001249) for the micro-manager repo). Thanks for minimizing spurious changes in the meantime.

wxs:
Great, thanks. I was wondering why others had not seen this issue, Windows/Linux seems to explain it. Do you have a Zeiss to test on? I'm not set up to build right now on Windows but could in theory support using our motorized stage if I figured that out.

marktsuchida:
Sorry for the lack of reply. I'll definitely get to this this week.

marktsuchida:
I pushed a commit to this branch to fix a couple of MSVC compiler warnings.
Otherwise, all the changes look safe on Windows, because `ZeissLong` and Windows `long` are the same anyway.
I don't currently have access to a Zeiss scope to test this on; if you have a Windows computer and can test the next nightly build, that would be great (but I think the changes of it breaking are low).
Thanks for the fix!

PR #237, opened 2022-08-22
Spinnaker: Binning returns 1 for Video Mode cameras

elliotsteele:
See : https://forum.image.sc/t/flir-chameleon3-camera-problem-with-transpose-and-mda/69955/9


marktsuchida:
Thanks, looks mostly good to me. To avoid future confusion, it may be best to add a comment to the `OnBinningEnum` function (or rename it to `OnBinningOrVideoModeEnum`, or make it two separate functions) to indicate that it is used for Video Mode as well as Binning.

elliotsteele:
Hi Mark, good point. I changed it to OnVideoMode since "Video Mode" cameras have a read only binnning property and cameras with Integer binning settings already use the separate "OnBinningInt" function

PR #238, opened 2022-08-24
Fix bug in TriggerScope DAC voltage when range goes negative

ieivanov:


marktsuchida:
Sorry for the slow reaction. Thanks for the fix!

PR #253, opened 2022-10-05
Device Adapter for LightSheetManager

bls337:
This is the device adapter for the LightSheetManager plugin, it maps physical devices to their logical names in the plugin.


marktsuchida:
I rebased and added to solution. (If you make changes in the future, please start a new branch off `main`.)

Issue #273, opened 2022-10-22
MMTime woes


marktsuchida:
[`MM::MMTime`](https://github.com/micro-manager/mmCoreAndDevices/blob/172de90f682288741837dd8bc7182ed386566295/MMDevice/MMDevice.h#L97) is a data type we use in device adapters.

Although the data type (which consists of `long sec_` and `long uSec_`) is defined in MMDevice, getting the "current time" as an `MMTime` value is done through CoreCallback (`GetCurrentMMTime()`) (which turns out to be a good thing).

The [current implementation in MMCore](https://github.com/micro-manager/mmCoreAndDevices/blob/172de90f682288741837dd8bc7182ed386566295/MMCore/CoreUtils.h#L105-L117) (which hasn't changed since the switch from ACE to Boost, some time between 2008 and 2010) returns (according to my hopefully correct testing) seconds since 2000-01-01 00:00:00 UTC, plus the current time zone offset. That is, if you are on a UTC-500 timezone (US Central Daylight Time), the result is (number of seconds since 2000-01-01 UTC) minus 18000 (= 3600 x 5).

The vast majority, if not all, of existing usage of `MMTime` by device adapters is for the purpose of computing time deltas (i.e. measuring time between events). For such purposes, it makes more sense to use a steady (monotonic) clock (such as `std::chrono::steady_clock`) rather than the wall clock time (time mapped to calendar date and time and subject to adjustment).

The problems with the current implementation are:
- Because of the use of wall clock time, time deltas may be affected by system clock adjustments (probably a minor, or at least hard-to-notice, issue under common conditions). We should use a steady clock.
- Because of the local time offset being part of the value, time delta behavior may be completely broken if run while the time zone changes, e.g., for daylight saving time (a major issue, although obviously not encountered very often -- but who knows?). We should use UTC or a steady clock.
- The current behavior is hard to reproduce except by keeping the exact same Boost-based code.

A theoretical fix would be to use `std::chrono::steady_clock` everywhere in device adapters where time deltas need to be computed. But there is no mechanical conversion, and it is not practical to make such a change without the risk of introducing hard-to-test bugs into device adapters (not to mention the huge amount of manual work needed). So we need a fix that preserves `MMTime` and its relevant behavior.

[...]

PR #274, opened 2022-10-23
MMCore: Replace most uses of Boost.Date_Time with std::chrono

marktsuchida:
Convert all but `MM::Time` (`GetCurrentMMTime()`) implementation, which requires resolving #273.

Setting as draft until I finish testing.


marktsuchida:
Fixed log date-time format to match previous.

PR #279, opened 2022-10-26
Remove StartTime-ms metadata tag from cameras

marktsuchida:
In most cameras, this tag contained the value of `GetCurrentMMTime()` at the beginning of sequence acquisition. That value is milliseconds since 2000-01-01 UTC, plus an offset dependent on the current time zone.

The utility of this tag is therefore not clear: it is not documented anywhere, its interpretation is complicated, and there is no evidence that it has ever been used in MMStudio or its plugins.

If we ever find a need for recording the time of start of acquisition, it is better to add a tag in MMCore so that all cameras have it, not just the ones that happened to copy-paste program this tag.


marktsuchida:
See [this previous comment](https://github.com/micro-manager/mmCoreAndDevices/issues/273#issuecomment-1287947155) for some more details.

PR #287, opened 2022-10-29
MMCore: Deprecate user id/host name/MAC functions

marktsuchida:
These never really belonged in MMCore (they are only there because it was a convenient place to get C++ functions wrapped for Java) and these days there are better alternatives in Java and Python. Removing these has a small benefit of simplifying the build config on macOS (and potentially Windows), aside from removing code that need maintenance (the macOS implementation currently gets a deprecation warning).

### DIFF
--- a/MMCore/MMCore.h
+++ b/MMCore/MMCore.h
@@ -629,9 +629,9 @@ public:
 
    /** \name Miscellaneous. */
    ///@{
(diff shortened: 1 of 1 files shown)


marktsuchida:
Setting to draft for now because it turns out `[[deprecated]]` doesn't work with SWIG (3.x at least).

Issue #292, opened 2022-11-02
MicroFPGA build


jdeschamps:
Hi!
I noticed that [MicroFPGA](https://github.com/micro-manager/mmCoreAndDevices/tree/main/DeviceAdapters/MicroFPGA) is not shipped with 2.0.0 or the NB. It is not in `micromanager.sln`.

Any specific reason? Should I make a PR to add it to the `.sln`file?

Thanks!

marktsuchida:
I was wondering about that (#207). Please feel free to make a PR.

jdeschamps:
Done https://github.com/micro-manager/mmCoreAndDevices/pull/293!

Issue #207, opened 2022-05-26
Projects not in micromanager.sln


marktsuchida:
Does anybody know why the following projects are not in micromanager.sln?
- [x] MicroFPGA (missing `.vcxproj.filters`; project needs `Win32` removed) #292
- [ ] MPBLaser (missing `.vcxproj.filters`; project needs `Win32` removed)

I found these by chance while looking at micro-manager/micro-manager.github.io#104. Want to make sure it's not a mistake.

marktsuchida:
In case this comes up in the future, here is how to list all projects that are not in micromanager.sln:
```sh
grep '^Project' micromanager.sln |cut -d , -f 2 |cut -d '"' -f 2 |tr \\ / |sort >in-sln
find . -name '*.vcxproj' |sed 's/^\.\///' |sort >not-in-sln
diff in-sln not-in-sln |grep '^>'
```

jdeschamps:
Are the filters necessary? I have seen that some other device adapters don't have `vcxproj.filters` (including some of mine).

marktsuchida:
Yes, the filters should be version controlled (unlike `.vcxproj.user`). The build works without them but I think the files don't display nicely in VS when they are missing. So adding the missing ones would be nice but not super urgent.

PR #293, opened 2022-11-06
Fix Mufpga project files

jdeschamps:
Related to 
https://github.com/micro-manager/mmCoreAndDevices/issues/207
https://github.com/micro-manager/mmCoreAndDevices/issues/292

- Removed Win32 builds
- Set PlatformToolset to v142
- Add filters


marktsuchida:
Thanks!

PR #301, opened 2022-11-17
NI100X minor cleanup

marktsuchida:
This happened as a side effect of investigating a crash (which may not to relate to NI100X as it turns out).

It does contain fixes for what are technically bugs (memory leak; `new[]` not paired with `delete[]`; possible(?) dereferencing of uninitialized pointers), although I'm not sure if any were observable bugs.

(Why am I still using `NI100X` instead of `NIDAQ`? It's a long story.)


joaomamede:
Hi Mark,

Can you do sequencing with the digital NI100X? I believe thanks to your cleanup I can now use 32bit unsigned integer where before the configuration file would crash if we would pass 8bits.

 I can't use the new NIDAQ module because one of my most important LED lines is at the last line of it and that bars me from using NIDAQ, that and the fact that I can't use PFI0 directly has a trigger input.

PR #304, opened 2022-11-23
89North LDI: handle unsolicited ok.

nicost:
Also removed a statement that always was true.  Not quite sure why there is a timeout in the loop reading the response (even though the serial port already has its own timeout), but leave that to avoid unexpected side effects.

### DIFF
--- a/DeviceAdapters/89NorthLDI/LDI.cpp
+++ b/DeviceAdapters/89NorthLDI/LDI.cpp
@@ -356,6 +356,7 @@ bool LDI::sendCOMMessage(const std::string & message, std::string& response)
 		return false;
 	}
 
+	response = "";
 	clock_t start = clock();
 	while (true)
 	{
--- a/DeviceAdapters/89NorthLDI/LDI.cpp
+++ b/DeviceAdapters/89NorthLDI/LDI.cpp
@@ -356,10 +356,8 @@ bool LDI::sendCOMMessage(const std::string & message, std::string& response)
 		return false;
 	}
 
-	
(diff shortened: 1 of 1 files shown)


marktsuchida (commented):
I don't think this is related to the issue you are addressing here, but there are multiple problems with this loop:
- `GetSerialAnswer()` already has its own timeout, so it is not clear what this loop is supposed to achieve.
- The check for `response != ""` is only "correct" if the caller of `sendCOMMessage()` passed an empty `response`, which is a precarious assumption.
- Instead, the return value of `GetSerialAnswer()` should not be checked.
- `clock_t` and [`clock()`](https://en.cppreference.com/w/cpp/chrono/c/clock) give CPU time, which doesn't make sense here (but, again, maybe the loop itself is not needed at all).

Just thought I'd point this out because it might be an opportunity to clean up while you are already set up to test this. Not a requirement for merging this.

In addition, I'm wondering if the code should guard against the case where multiple unsolicited `ok` messages have been received. I don't know if that is a possibility.

nicost:
Elliot Steele was afraid of the device regularly sending empty responses, hence the loop (with its own timeout).  

> The check for response != "" is only "correct" if the caller of sendCOMMessage() passed an empty response, which is a precarious assumption.

Good point!  I now explicitly set `response `to `"" ` before the loop, to assure "correctness".

I don't think it is needed to guard against multiple "ok"s.  Apparently, the LDI documentation says that it only replies to commands send, and it is kind of understandable where this one unsolicited "ok" comes from, so I'd rather not complicate the code further.

PR #307, opened 2022-12-05
Add State Sequencing to DStateDevice

ieivanov:
This PR adds the ability to sequence the State property of the `DStateDevice`. This is helpful in testing the channel sequencing capabilities of the acquisition engine as in https://github.com/micro-manager/pycro-manager/pull/458.

Here I also remove the Delay capability of the `DStateDevice`, which overlaps in functionality with `DWheel`.

In a future PR, I plan to rename the device from `DemoCamera` to `DemoHub` and separate the device across individual `.cpp` files.

### DIFF
--- a/DeviceAdapters/DemoCamera/DemoCamera.cpp
+++ b/DeviceAdapters/DemoCamera/DemoCamera.cpp
@@ -2822,6 +2822,7 @@ CDemoStateDevice::CDemoStateDevice() :
 numPatterns_(50),
 numPos_(10), 
 initialized_(false),
+changedTime_(0.0),
 busy_(false),
 sequenceOn_(true),
 gateOpen_(true),
@@ -2830,6 +2831,7 @@ isClosed_(true)
 {
    InitializeDefaultErrorMessages();
    SetErrorText(ERR_UNKNOWN_POSITION, "Requested position not available in this device");
+   EnableDelay(); // signals that the dealy setting will be used
 
    // Number of positions
    // -----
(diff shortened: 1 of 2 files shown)


marktsuchida:
Thanks, having sequencing capability is nice.
(If you are doing this for testing, you might also be interested in the SequenceTester device adapter, which I added to the build recently. It supports triggered sequencing and can show you what it is doing. Unfortunately it lacks documentation at the moment.)

> Here I also remove the Delay capability of the `DStateDevice`, which overlaps in functionality with `DWheel`.

I think it would be better to keep the Delay. People use demo devices in various ways to supplement real devices, and changing the behavior might break some (admittedly kludgy but working) configurations. Let's keep it, unless there is a very strong reason why it cannot be kept (I don't think it conflicts with sequencing, does it?).

> In a future PR, I plan to rename the device from `DemoCamera` to `DemoHub` and separate the device across individual `.cpp` files.

This would break hardware configurations on many people's microscopes. Please don't rename devices and device adapters.

marktsuchida:
Also better to default to sequencing off, for backward compatibility.

ieivanov:
>If you are doing this for testing, you might also be interested in the SequenceTester device adapter, which I added to the build recently. It supports triggered sequencing and can show you what it is doing.

I didn't know about the sequence tester, I'll have a look through it, thanks!

>I think it would be better to keep the Delay.

Got it, I'll add back the delay. I don't think that would interfere with the sequencing capabilities.

>This would break hardware configurations on many people's microscopes. Please don't rename devices and device adapters.

Sounds good, I'll keep the `DemoCamera` name. I think it will still be useful to separate the devices across individual `.cpp` files for easier editing, this will have no functional effect.

ieivanov:
@marktsuchida I think this is ready to merge, let me know if you have any other comments.

marktsuchida:
Thanks for the updates.

> I think it will still be useful to separate the devices across individual `.cpp` files for easier editing, this will have no functional effect.

I agree with this! DemoCamera could use quite a bit of cleanup in general.

ieivanov:
>I agree with this! DemoCamera could use quite a bit of cleanup in general.

Sounds good! I'll create a separate issue for this so we can keep track of it

PR #314, opened 2023-01-17
PVCAM: Added missing centroids parameters

tomhanak:
This PR adds parameters `PARAM_CENTROIDS_MODE`, `PARAM_CENTROIDS_BG_COUNT` and `PARAM_CENTROIDS_THRESHOLD` that were not added with initial centroids implementation a couple of years ago.

It should also fix a possible issue when the Photometrics camera (e.g. Prime-95B) was used in another application in Track Centroids mode and then open in Micro-Manager that understood Locate mode only and was unable to change the mode.

The camera can provide "extended metadata" in Centroids modes that is now added image tags sent to MM core.

PR #317, opened 2023-02-01
PVCAM: Added new parameters describing image format

tomhanak:
The recent PVCAM library is able to do local post-processing of the image coming from the camera. For example it is a fully transparent decompression of bit-packed frames for some USB cameras or frame rotation/flip.

However, there are also new features that could change the image format sent from PVCAM to the application, e.g. frame summing (pull request will land here soon) can be configured to sum 20 frames with bit-depth 12 and generate one output frame with 16-bit pixel.

This PR extends existing code to use new PVCAM parameters describing the output image format and fallback to previously used parameters. Additionally, the dependencies among various parameters were at least documented and hopefully most of them handled in the code to refresh the cached values and UI.

This PR requires new PVCAM SDK files added to 3rdpartypublic repository rev. 210 to build without errors.


marktsuchida:
Thanks for these detailed descriptions!
We're already building with 3rdpartypublic r210 (see the `3rdpartypublic-revision` file in the main `micro-manager` repository), since I realized that #314 required it to build.

PR #318, opened 2023-02-02
PVCAM: Added new parameters for SW frame summing

tomhanak:
There are only a few cameras that offer frame summing post-processing feature done directly on HW. Latest PVCAM library introduced new set of parameters that add frame summing feature for all cameras.

That feature can be turned on and off, the number of frames to sum is configurable, and also the output image format can be changed. For now the output bit depth can be either 16- or 32-bits. 24-bit output might be added later. However, Micro-Manager doesn't support images with more than 16 bits per pixel in single channel, so the PVCAM adapter shows an error message to the user and reverts the settings back after confirmation.


tomhanak:
This is the last change set. More will come in a few weeks/months, mainly cleanup, legacy code removal or switching from boost to C++11/14 features. Thanks for fast merges.

PR #321, opened 2023-02-09
add SigmaKoki Adapter

ABEDToufikSK:
SigmaKoki device adapter/open source

### DIFF
--- a/DeviceAdapters/SigmaKoki/Camera.cpp
+++ b/DeviceAdapters/SigmaKoki/Camera.cpp
@@ -950,7 +950,6 @@ int Camera::InsertImage()
 	// Important:  metadata about the image are generated here:
 	Metadata md;
 	md.put("Camera", label);
-	md.put(MM::g_Keyword_Metadata_StartTime, CDeviceUtils::ConvertToString(sequenceStartTime_.getMsec()));
 	md.put(MM::g_Keyword_Elapsed_Time_ms, CDeviceUtils::ConvertToString((timeStamp - sequenceStartTime_).getMsec()));
 	md.put(MM::g_Keyword_Metadata_ImageNumber, CDeviceUtils::ConvertToString(imageCounter_));
 
@@ -1062,9 +1061,9 @@ void MySequenceThread::Start(long numImages, double intervalMs)
 	stop_ = false;
 	suspend_ = false;
 	activate();
-	actualDuration_ = 0;
+	actualDuration_ = MM::MMTime{};
 	startTime_ = camera_->GetCurrentMMTime();
-	lastFrameTime_ = 0;
+	lastFrameTime_ = MM::MMTime{};
 }
 
 bool MySequenceThread::IsStopped() {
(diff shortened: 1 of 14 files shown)


marktsuchida:
Thanks, working on this now.

marktsuchida:
@ABEDToufikSK This is building on my local machine after a few fixes. I pushed a few commits to your branch:
- Removed the `StCamD.lib` and `StCamD.h` files. We prefer to leave binaries and third-party SDK files out of our source tree as much as possible (for size and license reasons); instead I will set up our automated build to use the header and libraries from Sentech's StCamUSBPack 220929 (from sentech.co.jp).
- Some fixes to `MM::MMTime` usage so that it compiles with recent changes to MMDevice (#278 and #279).
- Standardized various VisualStudio project settings and removed 32-bit configurations.
- Removed the old SigmaKoki (in SecretDeviceAdapters) from the Windows build and add this new one.

For the Sentech SDK, it looks like you were using a version from 2015. Since this is a simple C interface and the latest version only adds functions (nothing changed or removed), I expect that using the latest (220929) SDK will make no difference. But let me know if this is a problem.

As soon as I have your confirmation on that, I can go ahead with the merge.

The following is a note to myself for what remains to be done before merging:
- Squash the commit removing SDK files so that they do not appear in mainline history
- Add Sentech SDK files to 3rdparty and update 3rdparty version in micro-manager repo
- Remove the old SigmaKoki adapter from SecretDeviceAdapters macOS build (the new one is Windows only for now, but best to avoid having different versions; if we want to re-add macOS/Linux support, we can do so by disabling the camera using the preprocessor)

ABEDToufikSK:
@marktsuchida 
_ We understand about removing the StCamD.lib and StCamD.h files.
_ Using the latest (220929) SDK is ok. We already removed the oldest and installed the latest. 
_ About Some fixes to MM::MMTime usage etc. :  we wanted to check if camera still working even with change.
after squashing the commit, we were able to release sigmakoki dll, but when running in my local MM(2.0) (by debugging Java) the dll wont appear in hardware configuration, looks like the standardization VS project settings commit made SK adapter not appearing. Do you have same issue from your side or I missed something here. as I'm using ver [Windows 64-bit](https://download.micro-manager.org/release/2.0/Windows/MMSetup_64bit_2.0.0.exe) 

Let us know.

marktsuchida:
@ABEDToufikSK The changes to `MMTime` (October, 2022) made the device interface incompatible, so we incremented the device interface version from 70 to 71. Micro-Manager checks this version when loading a device adapter, to avoid talking to a binary-incompatible module. So you need a recent Micro-Manager installation that uses device interface version 71: it is best to install the latest nightly build to test.

You can find the device interface version used by a given MM installation by going to Help > About. When a device adapter fails to load due to incompatible device interface version, this fact should be logged in the CoreLog.

The reason it used to work for you is most likely that both your MMDevice source code and MM installation were at device interface version 70 (or older but in any case matching). But building the tip of this pull request's branch will use the latest MMDevice and build for version 71.

I'll wait for you to test before merging this.

ABEDToufikSK:
@marktsuchida, 
Thank you for making this clear, it was all about a version as I forked the repository more than half year. 
The adapter works like charm now! it's ok to merge.  Thanks in advance.
Let me know if anything, we will soon update the documentation on the public website: [SigmaKoki](https://micro-manager.org/SigmaKoki)
Regards,

PR #324, opened 2023-02-15
bump basler sdk version

henrypinkard:


marktsuchida:
Together with the "unclear debugging code", it looks like you also removed a call to `ClearROI()`. Is that intentional?

Other than that, LGTM.

PR #327, opened 2023-02-17
MMCore: Cosmetic changes to thread pool and task classes

tomhanak:
The ```ThreadPool``` and related task classes were originally developed as a part of PVCAM device adapter and later copied to MMCore where the same functionality is needed. So far these classes are not considered a part of public API of MMCore, thus we have two copies.

This PR basically synchronizes the implementation with changes done in PVCAM adapter, namely using ```std::unique_ptr``` to store thread instances (that addressed a TODO note), restricting the task types to derive from ```Task``` base class, used C++11 keywords like ```override``` or ```final```.
The two copies of each file now differ only in one line in the copyright header comment stating the actual module name the code belongs to.

### DIFF
--- a/MMCore/Semaphore.cpp
+++ b/MMCore/Semaphore.cpp
@@ -26,7 +26,6 @@
 #include <mutex>
 
 Semaphore::Semaphore()
-    : count_(0)
 {
 }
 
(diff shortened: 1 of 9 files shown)


marktsuchida:
Thanks for keeping this maintained!

PR #331, opened 2023-03-03
Add support for TGGALVO as scanner device

bls337:


henrypinkard:
@marktsuchida When this type of thing happens, does it mean that CI breaks for everything in micro-manager/micro-manager? If so, maybe it is worth switching to the PR-based workflow I made for pycro-manager? Happy to take a look at it if you think it would help

marktsuchida:
@henrypinkard Yes, the CI breaks for everything. I assume you mean PR-based in the sense that `micro-manager` will keep using the existing `mmCoreAndDevices` until a human approves the update? Yes, I still intend to do that! (I would like to study what you did with the separate app, since I didn't have a chance to see that before you enabled it.)

In practice these build breakages are relatively infrequent (I often test the build locally before merging when the code is from less frequent contributors, 100% of the time in the case of new device adapters). I would have added a (possibly temporary) fix to the build if necessary, but in this case no other changes were happening so I didn't feel the need to rescue the nightly build. Better have @bls337 do the fix in case there is some code intent that I might miss if I do the minimal (two-character) fix needed to compile.

henrypinkard:
> I assume you mean PR-based in the sense that `micro-manager` will keep using the existing `mmCoreAndDevices` until a human approves the update?

Yeah, that or with automerge, maybe with some select rules. Just replacing the push directly to main strategy so that some tests automatically run.

> (I would like to study what you did with the separate app, since I didn't have a chance to see that before you enabled it.)

Sounds good, let me know if you have any questions!

bls337:
Sorry about that, I will be more careful in the future.

marktsuchida:
@bls337 No problem at all! The problem is really with our current setup that doesn't check the build before merging. And as I wrote above, it's not blocking anybody else's changes at the moment (or I would have added a temporary fix, easily).

Just to be clear, this hasn't been fixed yet. Can you create a separate PR with a fix? (It compiles with `hub` -> `hub_` and adding a `)` at the end of that line, but I just wanted to confirm that the file is otherwise the version you intended.)

PR #333, opened 2023-03-08
Update PI_GCS_2 device adapter

marktsuchida:
From Steffen Rau, PI:

We did some internal changes to simplify code and added:

- Support for upcoming devices with GCS 3.0 command syntax
- Support TCP/IP with PI_GCS2_DLL
- Support of joysticks connected to controller with “HIN” command

(Changes provided via email and committed by Mark Tsuchida.)


marktsuchida:
Waiting to hear from Steffen on a possible issue, so don't merge this just yet.

PR #334, opened 2023-03-10
Add Lumenera infinity device adapter

henrypinkard:


marktsuchida:
Looks good to me (checked mostly the build files, by eye).

Do you mind if I force-push a linearized and rebased version of your history? It will preserve all your individual commits except for the ones merging from `main`, and the code in `DeviceAdapters/Lumenera` will not be affected. Mostly so that we get a readable history, e.g., in `gitk`.

henrypinkard:
Thanks for checking! Sure, go ahead with the force push.

marktsuchida:
Done.

PR #353, opened 2023-06-19
MicroFPGA: update

jdeschamps:
Hi!

A small update of MicroFPGA device adapter:
- Change error message to include most common problem (wrong baudrate)
- Change (seemingly...) the default baudrate 

Now a small question before merging: I tested the new version of the adapter and the [default value of the baudrate](https://github.com/jdeschamps/mmCoreAndDevices/blob/afcad44dddeb3cef1a1a2accd75a997f29941a3e/DeviceAdapters/MicroFPGA/MicroFPGA.cpp#L194) is not effective. I still get "9600" when using the hardware configuration wizard. Do you have any idea why?

I can also open an issue or post on the forum.

### DIFF
--- a/DeviceAdapters/MicroFPGA/MicroFPGA.cpp
+++ b/DeviceAdapters/MicroFPGA/MicroFPGA.cpp
@@ -145,7 +145,7 @@ MicroFPGAHub::MicroFPGAHub() :
 	SetErrorText(ERR_PORT_OPEN_FAILED, "Failed to open MicroFPGA USB device.");
 	SetErrorText(ERR_BOARD_NOT_FOUND, "Did not find an MicroFPGA board with the correct firmware. Is the MicroFPGA board connected to port?");
 	SetErrorText(ERR_NO_PORT_SET, "Hub Device not found. The MicroFPGA Hub device is needed to create this device.");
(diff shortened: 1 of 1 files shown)


marktsuchida:
> I tested the new version of the adapter and the [default value of the baudrate](https://github.com/jdeschamps/mmCoreAndDevices/blob/afcad44dddeb3cef1a1a2accd75a997f29941a3e/DeviceAdapters/MicroFPGA/MicroFPGA.cpp#L194) is not effective. I still get "9600" when using the hardware configuration wizard. Do you have any idea why?

The line you link to is what is used when clicking on the "Scan Ports" button while configuring the hub. The baud rate setting should then be preserved if communication succeeds with one of the ports. If not, the port settings revert back to their original value. If the port scanning fails ("No valid ports found" displayed under the "Scan Ports" button), you'll probably see 9600.

If that is not how it is behaving, it might be a bug.

There is currently no way for devices to set the port settings other than via a successful port scan.

jdeschamps:
Ah I didn't know that. Thanks!

The "scan port" button is actually not available with this device adapter hub. There is no communication until I select the baud-rate and click  "OK" in the wizard.

PR #355, opened 2023-07-11
Updates pom.xml to version 10.4.0.  

nicost:
In the future, the MMCore.cpp and pom.xml (in MMcoreJ) should be kept at the same value. It would be nice if this could be automated.  Ready to be pushed to maven on a system that has the PGP keys installed correctly.


henrypinkard:
FYI the maven side of this can be automated with something like [this](https://github.com/micro-manager/pycro-manager/blob/main/.github/workflows/build_and_deploy.yml). You just have to add the appropriate keys to the repository secrets

PR #359, opened 2023-07-24
Adding ZML and Objective Changer

martinzak-zaber:
In our PR, we are switching our plugin to our SDK which was developed in the past few years. We would like to get access to SVN to commit binaries (before this gets merged in). Our SDK uses a mechanism that allows for sharing of serial ports among multiple programs. Therefore, we are opting out of serial port management by MM.

Additionally, we are adding a new device called Objective Changer.

It's worth noting that our knowledge of autotools is rather limited.

We are looking forward to hearing your comments and suggestion.

Thank you for your continuous support.

Zaber Software Team


marktsuchida:
Thank you for the update.

We are about to migrate our SVN repositories and cannot give new write access currently, but I'm happy to set up the files needed for the build. Will this build with the Zaber Motion Library, C++, from your website? (Kudos for making it open source!)

martinzak-zaber:
Yes, the binaries are the C++ ZML from the website, but it's a particular version, and all the platforms melted together. I have prepared instead a zip file that you can download and extract into 3rdpartypublic:

https://drive.google.com/file/d/1O9BvWhQi_jjcVJyiHPc5Qm6sfuwZjMQ1/view?usp=sharing

Please note that file contains symlinks.

Let us know if there is any trouble!

marktsuchida:
Apologies for the delay. I think this should work if users can install the DLLs/.so/.dylibs on their own (since they are ~30 MB, I'd rather not add them to the MM installer).

The symbolic links should be [okay](https://svnbook.red-bean.com/nightly/en/svn.tour.cycle.html#svn.tour.cycle.edit) since they are used for the Linux/macOS libraries only. I'll make sure to upload from macOS or Linux.

I should be able to work on this tomorrow. Thanks for your patience.

martinzak-zaber:
We understand the decision though we are disappointed by the inconvenience this brings to our customers.

I'll work on the PR to make the necessary changes.

marktsuchida:
Thank you for your understanding.

Just an idea, and I have no idea if it will work, but is it possible for the device adapter (mmgr_dal_Zaber.dll) to statically link to ZML and end up with a smaller total size? I was going to try this but it didn't seem that the ZML CMake script supports building as a static library.

Another idea would be to have an end-user installer for the ZML DLL. The simple version of this would require the user to copy the DLL(s) into the Micro-Manager folder. Sometimes it is possible to avoid this (by loading the DLL from a standard install location), but the feasibility may depend on various factors.

martinzak-zaber:
I've removed the build.xml file in 337adb7faa66c7a77e550728c5c79ca007e95e32 to prevent copying of the .dll . We will make our customers download the file and copy them over to the MM installation folder. Possibly even make an automated installer. That way, it won't interfere with any other installed version.

We cannot make it a static library, as far as I know.

Thank you for your help.

marktsuchida:
@martinzak-zaber Thanks for your patience. I pushed 2 commits to make this build in our CI:
- Change the 3rdpartypublic directory to include `Zaber/` (to match existing convention, and in case anything Zaber-related needs to be added in the future)
- Locally set C++ standard to C++17 for Zaber only in autotools build (since our overall build must remain C++14 for now)

martinzak-zaber:
Thank you for merging our update.

I was hoping C++17 wouldn't be necessary since we made a special release of v3 (the content of the tar.gz) of ZML that does not require C++17 yet. Either way, we welcome it, as we can now use our latest version for further updates.

tlambert03:
quick q about this PR.  I noticed that our ubuntu tests are [failing in napari-micromanager](https://github.com/pymmcore-plus/napari-micromanager/actions/runs/5992547650/job/16252045021?pr=280#step:7:7114) with:

```
Zaber.h:32:10: fatal error: zaber/motion/microscopy.h: No such file or directory
   32 | #include <zaber/motion/microscopy.h>
```

Do we need to update our build recipe to download the Zaber motion library now?  Or will that be added to the autogen setup in a followup PR?  [Currently](https://github.com/pymmcore-plus/napari-micromanager/blob/6abd0f50cf7840b7a08b28bd0ad9a428c484f284/.github/workflows/test.yml#L65-L89), it looks like:

```
          git clone https://github.com/micro-manager/micro-manager.git
          cd micro-manager
          git submodule update --init --recursive
          ./autogen.sh
          ./configure --without-java
          make -C mmCoreAndDevices/MMDevice -j$(nproc)
          make -C mmCoreAndDevices/MMCore -j$(nproc)
```

martinzak-zaber:
The includes as well as binaries were added to `3rdpartypublic` SVN repo. Did you perhaps forget to do SVN Update?

martinzak-zaber:
I see that this particular file is missing. That's a mistake on our side. Let me create another merge request to fix that.

martinzak-zaber:
Sorry for the confusion. The file is there. Just SVN update is needed.

marktsuchida:
@tlambert03 This is a regression, because it looks like you were able to build the C++ code without 3rdpartypublic (at least on non-Windows). That's probably a good capability to keep. I'll look at fixing this.

marktsuchida:
No problem. I'll do this later today.

marktsuchida:
Replaced with the new zaber-motion.tar.gz in 3rdpartypublic r212.
It looks like `svn add` ignored the linux shared libs by default; this is also fixed in r212.

This did indeed remove the requirement for C++17 (checked macOS, Windows), so undid the settings in #370.

(I still need to look into #368 separately.)

martinzak-zaber:
Thank you 🙏

Issue #363, opened 2023-08-11
Point Grey Device Adapter and MM version conflicts


kmdouglass:
Hi everyone,
We're having issues with the Point Grey device adapter in Micro-Manager versions newer than February 2, 2022. On this date, the FlyCap SDK was updated from 2.10.3.266 to 2.13.3.61 as described in #130 . We noticed that the device adapter will not load in these newer MM versions if FlyCap 2.10.3.266 is installed.

So, we tried instead to install FlyCap 2.13.3.61 but now we cannot proceed past the hardware configuration wizard because the device adapter code returns an error if the FlyCap DLL version is not 2.10.3.266 in the `Initialize` method: https://github.com/micro-manager/mmCoreAndDevices/blob/main/DeviceAdapters/PointGrey/PointGrey.cpp#L263-L268

I was wondering whether you can confirm or not that updating the check on the required version of the DLL in `PointGrey.cpp` will fix the issue. I could set up a Windows compilation environment to fix it, but I wanted to run it by someone here first to make sure it's the right fix.

Any ideas? Thanks!

marktsuchida:
Fixed by #364.

PR #364, opened 2023-08-15
Update FlyCap2 version check in Point Grey device adapter

kmdouglass:
This change addresses the problem with the Point Grey device adapter that is described in #363 . MM is currently compiled against FlyCap 2.13.3.61, but a version check in the device adapter code requires a DLL from version 2.10.3.266. This mismatch renders the current device adapter code unusable.

I verified that this change works on a Flir Grasshopper3 U3-23S6M.


marktsuchida:
Thank you so much for working this out!

PR #367, opened 2023-08-25
Remove FixSnprintf.h

marktsuchida:
This file was there to allow the code to compile with either Visual Studio 2010 or VS2015+.
It is no longer necessary or useful (we don't support compiling with anything older than VS2019), and currently has no effect.

Since people keep copy-pasting the inclusion of this file (even where `snprintf()` is not used!), let's get rid of it for good.

There is also often an adjacent `#include <Windows.h>` that is cargo culted in many cases, but that is harder to mechanically remove because sometimes it is actually needed.

No change to device interface.


henrypinkard:
Great!

PR #370, opened 2023-08-30
Zaber: Revert C++17 requirement

marktsuchida:
With corrected ZML on 3rdpartypublic, C++17 is no longer required.

See https://github.com/micro-manager/mmCoreAndDevices/pull/359#issuecomment-1696564608

### DIFF
--- a/DeviceAdapters/Zaber/Makefile.am
+++ b/DeviceAdapters/Zaber/Makefile.am
@@ -1,7 +1,4 @@
-# We require C++17 because libzml is C++17. Setting the C++ standard for
-# individual device adapters is a hack and should not be imitated except when
-# there is no other option.
-AM_CXXFLAGS = $(MMDEVAPI_CXXFLAGS) -I$(MMTHIRDPARTYPUBLIC)/Zaber/zaber-motion/include -std=c++17 -Wno-dynamic-exception-spec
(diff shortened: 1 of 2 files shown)


martinzak-zaber:
Hello @marktsuchida,

Hope you are doing well. We are considering some updates to our integration in MicroManager. We wanted to ask whether C++17 support is possible or perhaps coming soon? 

The answer determines which of the following options we'll hope to submit next:

1. Patch update to the old version of our library (3.4.3). Only introduce support for our new device variation.
2. Major update with recent version (6.6.0). Introduce support for the autofocus feature of our microscopes.

The newer version of the library is firmly C++17 so that's why we're asking. Can you please advise?

Thank you.

marktsuchida:
@martinzak-zaber Thanks for asking. I am pretty sure we can accommodate C++17 for particular device adapters. If I remember correctly, the changes that are reverted in this PR did work (they were reverted because it turned out that you were not actually using C++17 at the time).

If you make the same changes to the build scripts to enable C++17 and confirm that it works for the Zaber adapter, then I don't see any problem.

You will still need the `-Wno-dynamic-exception-spec` to work around our MMDevice headers.

martinzak-zaber:
Will do that. We much appreciate your support, @marktsuchida.

PR #371, opened 2023-08-30
Zaber: Skip build when ZML not found

### LINKED ISSUE #368: Linux C++ (`--without-java`) should build without `3rdpartypublic`
Regression caused by #359.

I'll look at fixing this; it will require adding dependency detection for ZML to `configure.ac` instead of hard-coding in `Makefile.am`, so that `Zaber` is not built when ZML is not present.

marktsuchida:
Closes #368.

Without this, 3rdpartypublic is required for the Mac/Linux build. We want to keep it so that only 3rdpartypublic/classext is required on those platforms (and not even that with `--without-java`).

This change also removes the special rpath flags. In the official macOS build, dependency paths are handled by a spearate script. On Linux, the only currently supported way is to build on the system where MM will be run, so (I hope) it is not necessary to use a special flag.

### DIFF
--- a/DeviceAdapters/Makefile.am
+++ b/DeviceAdapters/Makefile.am
@@ -65,6 +65,9 @@ endif
 if BUILD_V4L2
    V4L = Video4Linux
 endif
+if BUILD_ZABER
+   ZABER = Zaber
+endif
 
 # Please keep these ASCII-lexically sorted (pass through sort(1)).
 SUBDIRS = \
@@ -90,6 +93,7 @@ SUBDIRS = \
 	$(SPOT) \
 	$(USBMANAGER) \
 	$(V4L) \
+	$(ZABER) \
 	AAAOTF \
 	AOTF \
 	ASIFW1000 \
@@ -203,7 +207,6 @@ SUBDIRS = \
 	Xcite \
 	YodnE600 \
 	Yokogawa \
-	Zaber \
 	ZeissCAN \
 	ZeissCAN29 \
 	kdv \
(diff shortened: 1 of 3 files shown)


marktsuchida:
@martinzak-zaber Please let me know if the results from this on Linux are not satisfactory (I did check that it builds).

martinzak-zaber:
> @martinzak-zaber Please let me know if the results from this on Linux are not satisfactory (I did check that it builds).

Hey @marktsuchida. It does not work, unfortunately.

```
RPATH_FLAGS = -Wl,-rpath -Wl,\$$ORIGIN
dist_deviceadapter_DATA = \
	$(ZML_DIR_LIB)/libzml.so.3.4 \
	$(ZML_DIR_LIB)/libzaber-motion-lib-linux-amd64.so.3.4.3
```

The `dist_deviceadapter_DATA` was ensuring that the library gets copied during `make install` along with the Zaber adapter. The RPATH_FLAGS was ensuring that the library gets loaded from the directory where the adapter is located. Without those lines, the user would have to manually copy the files into the standard lib directory `/usr/lib`.

We would prefer those lines to be returned rather than make the user do more manual steps to make our adapter work.

martinzak-zaber:
@marktsuchida I can perhaps prepare a merge request that keeps the changes you've made and does not break our device adapter. Hopefully, it can satisfy everyone.

marktsuchida:
Thanks, I'm already half-done doing what I think is similar to what you are proposing (restore the copy of libs on Linux only, only when Zaber build is enabled). I'll ask you to review the PR before merging.

marktsuchida:
Created #373.

PR #372, opened 2023-08-31
save pixel size settings in saveSystemConfiguration

tlambert03:
This PR would add PixelSizeConfigs to the output of `saveSystemConfiguration` (similar to MMStudio's output).

@marktsuchida, haven't tested this locally, so let me know if there's any CI to confirm that stuff works, or if this repo just relies on local testing / visual inspection

### DIFF
--- a/MMCore/MMCore.cpp
+++ b/MMCore/MMCore.cpp
@@ -7327,7 +7327,7 @@ void CMMCore::saveSystemConfiguration(const char* fileName) throw (CMMError)
       
       os << MM::g_CFGCommand_PixelSize_um << ',' << configs[j] << ',' << getPixelSizeUmByID(configs[j].c_str()) << endl;
 
-      std::vector<double> affine = getPixelSizeAffine(configs[j].c_str());
+      std::vector<double> affine = getPixelSizeAffineByID(configs[j].c_str());
       os << MM::g_CFGCommand_PixelSizeAffine << ',' << configs[j] << ',';
       for (size_t i = 0; i < affine.size(); ++i) {
          os << affine[i];


marktsuchida:
The sad news is that I am quite uncertain about the current state of `saveSystemConfiguration()`. It is never called by MMStudio, so effectively it hasn't been tested in years, and it's probably been nearly a decade since I last looked at it. I seem to remember that some things were missing, and maybe it was the pixel size configs. Were you able to confirm that the pixel size settings were the only thing missing?

There is no unit test for this function (although it would be nice if there were), so yes, the current practice is just local testing. I'll take a quick look at the code and test it before merging.

MMCore/MMCore.cpp
```diff
@@ -6700,6 +6700,34 @@ void CMMCore::saveSystemConfiguration(const char* fileName) throw (CMMError)
...
+      for (size_t k=0; k<c.size(); k++)
+      {
+         PropertySetting s = c.getSetting(k);
+         os << MM::g_CFGCommand_ConfigPixelSize << ',' << configs[j] 
+            << ',' << s.getDeviceLabel() << ',' << s.getPropertyName() << ',' << s.getPropertyValue() << endl;
+      }
+      
+      os << MM::g_CFGCommand_PixelSize_um << ',' << configs[j] << ',' << getPixelSizeUmByID(configs[j].c_str()) << endl;
+
+      std::vector<double> affine = getPixelSizeAffine(configs[j].c_str());
```
marktsuchida, line 6721: This is a C++ trap. `getPixelSizeAffine(bool)` is being called (with the non-null pointer interpreted as `true`); there is no `getPixelSizeAffine(const char *)`.

marktsuchida:
In MMStudio (which is effectively the reference implementation for config file writing), the affine transforms are only written when they are set. It would be best if we do the same here.

But the situation is a little murky. If you set the affine transform for even one pixel size preset in Pixel Size Calibration, all presets now have an affine transform saved in the config file. The unset ones get [1 0 0 0 1 0] (i.e., identity transform), which would make sense if the affine transform is applied _in addition_ to the conventional pixel size, but that is not how it works. Is there a deeper problem lurking here? @nicost.

tlambert03:
yeah, i noticed that too.  For what it's worth, I recently re-implemented all of this logic in pymmcore-plus including the [microscope model](https://github.com/pymmcore-plus/pymmcore-plus/blob/main/src/pymmcore_plus/model/_microscope.py) and [writing the config file from the model](https://github.com/pymmcore-plus/pymmcore-plus/blob/main/src/pymmcore_plus/model/_config_file.py) so I could create a full [config wizard GUI in pymmcore-widgets](https://github.com/pymmcore-plus/pymmcore-widgets/pull/183) ... and there I also [skip default affines](https://github.com/pymmcore-plus/pymmcore-plus/blob/b2a2f524d9a743c84b234dced9b0045f68551fe3/src/pymmcore_plus/model/_config_file.py#L163-L164) as in MMStudio

so it's not super critical that I have this change in mmCoreAndDevices.  But, seems like a natural place for it to be the reference implementation :)

MMCore/MMCore.cpp
marktsuchida: Note that it already compiles, just returns the wrong data. I think you can use `getPixelSizeAffineByID(configs[j].c_str())`.

marktsuchida:
> But, seems like a natural place for it to be the reference implementation :)

This I agree with.

MMCore/MMCore.cpp
tlambert03: thanks!

nicost:
> But the situation is a little murky. If you set the affine transform for even one pixel size preset in Pixel Size Calibration, all presets now have an affine transform saved in the config file. The unset ones get [1 0 0 0 1 0] (i.e., identity transform), which would make sense if the affine transform is applied _in addition_ to the conventional pixel size, but that is not how it works. Is there a deeper problem lurking here? @nicost.

My memory is vague.  I believe that I though it safer to store the identity matrix by default, rather than nothing, or an all zero matrix.  Probably no deep thought went into this....

marktsuchida:
I guess then I would be okay with merging something that behaves exactly the same way as the Java writer, for now. This for the cases: 1) no affine matrices set, 2) some presets have affine matrices set, and 3) all presets have affine matrices set (probably 2 and 3 are the same case).

It would seem that any code that wants to use the affine transform needs to treat [1 0 0 0 1 0] as a special "null" value indicating the absence of an affine transform. This probably does not cause logical problems in practice, because the only case where [1 0 0 0 1 0] should be treated as an actual affine matrix is when the pixel size is 1.0 um/px, but then the result is the same as having no affine matrix. Whew.

But it might be worth improving the Core API so that callers don't need to (know that they have to) manually check for the special value.

MMCore/MMCore.cpp
```diff
@@ -7309,6 +7309,34 @@ void CMMCore::saveSystemConfiguration(const char* fileName) throw (CMMError)
...
+
+      std::vector<double> affine = getPixelSizeAffineByID(configs[j].c_str());
+      os << MM::g_CFGCommand_PixelSizeAffine << ',' << configs[j] << ',';
+      for (size_t i = 0; i < affine.size(); ++i) {
+         os << affine[i];
+         if (i < affine.size() - 1) {
+            os << ',';
+         }
+      }
+      os << endl;
```
tlambert03, line 7338: this is the main remaining question.  cc @marktsuchida

PR #373, opened 2023-09-05
Zaber: Copy ZML libs during Linux install

marktsuchida:
For now this is probably the simplest way to make Zaber work when built under Linux. The copy only happens if Zaber is being built.

Also restore the setting of rpath to `$ORIGIN` in the Zaber device adapter.

@martinzak-zaber Can you check that this works for you?


martinzak-zaber:
> For now this is probably the simplest way to make Zaber work when built under Linux. The copy only happens if Zaber is being built.
> 
> Also restore the setting of rpath to `$ORIGIN` in the Zaber device adapter.
> 
> @martinzak-zaber Can you check that this works for you?

@marktsuchida Everything works on our side. Thank you.

PR #376, opened 2023-09-08
MMCore: Enforce pre/post initialization rules

### LINKED ISSUE #366: Prevent crash on illegal access to uninitialized devices
Add checks to DeviceInstance member functions; throw when device is uninitialized and action cannot be performed.
(Currently it may crash, throw an exception, do nothing, or who knows what.)

See micro-manager/pymmcore#84.

marktsuchida:
- Most non-property operations require an initialized device
- Pre-init properties must not be set after initialization
- Device initialiation may only be attempted once per device instance

These rules have been unstated previously, but had to be observed. Now make it an error (throw a `CMMError`).

All `CMMCore` member functions that do _not_ throw `CMMError` were reviewed to make sure newly-thrown exceptions are not left uncaught in SWIG wrappers.

MMCore version bumped 10.4.0 -> 10.5.0.

Closes #366.

Cc: @tlambert03

### DIFF
--- a/MMCore/Devices/AutoFocusInstance.cpp
+++ b/MMCore/Devices/AutoFocusInstance.cpp
@@ -22,13 +22,13 @@
 #include "AutoFocusInstance.h"
 
 
-int AutoFocusInstance::SetContinuousFocusing(bool state) { return GetImpl()->SetContinuousFocusing(state); }
-int AutoFocusInstance::GetContinuousFocusing(bool& state) { return GetImpl()->GetContinuousFocusing(state); }
-bool AutoFocusInstance::IsContinuousFocusLocked() { return GetImpl()->IsContinuousFocusLocked(); }
-int AutoFocusInstance::FullFocus() { return GetImpl()->FullFocus(); }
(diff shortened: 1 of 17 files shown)


tlambert03:
I like it!!

What are your thoughts on adding an isInitialized(label) method?  I see that with the current structure, it doesn't _really_ matter, since requireInitialized errors are all being caught by core calls and dropped... but it would allow "look-before-you-leap" style approaches too.  Unless there's a specific aversion to giving a public accessor to that data?

marktsuchida:
No reason to hide it, but it's a little tricky, because devices are actually in one of 3 states:
- uninitialized (`initializeDevice()` never called since loading)
- initialized (successfully on first try)
- initialization failed (device cannot be reused without unloading/loading)

Presumably `isDeviceInitialized()` would return true only for the second state, but then a false return value would not indicate "can initialize". I suppose we could have a function returning an enum with these 3 states.

(Technically there is also a "shut down" state, but (thankfully) this is not observable outside of MMCore.)

tlambert03:
I think the enum response would be great (similar to deviceDetectionStatus). Informative, extensible...

PR #408, opened 2023-11-29
MMCoreJ build: simplify mvn build

nicost:
No longer need to move TaggedImage.java before each and every build. Correct path to swig generated files in pom.xml.


marktsuchida (approved):
- [x] Build from MMCoreJ_wrap.vcxproj works correctly
- [x] Full build from Ant (`ant stage`) works correctly
- [x] Resulting MMCoreJ.jar has same contents as before

Looks good!

marktsuchida:
Oops, I forgot to test the macOS build, and it fails. Should be easy to fix, working on it now.

PR #412, opened 2023-12-09
Avoid 'using namespace' in MMDevice, MMCore, DemoCamera and other cleanup

marktsuchida:
`using namespace std` causes [nasty problems](https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice) when new C++ standards add symbols to the `std` namespace (the same applies to other namespaces).

It's hard to immediately do anything about the many device adapters that do this (due to the complexity of C++, there is no automatic way to do an 100% safe fix, as far as I know), but let's set good examples in our core code.

This PR also applies `#pragma once` consistently to headers in these projects, replacing error-prone include guards.

Also, `endl` in MMCore has been replaced by `'\n'` rather than `std::endl`. `std::endl` is exactly equivalent to `'\n'` plus flushing the stream, but none of these uses need a flush.

No change to the device (binary) interface or the MMCore API. One `#include` needed to be added to Andor, due to cleaning up includes in MMDevice/ImgBuffer.h.


marktsuchida:
Note to self:

- [ ] Check macOS build before merging
- [ ] Check Linux build before merging

marktsuchida:
It turns out that DemoCamera uses unqualified `min` and `max`, and under non-Windows this was previously picking up `std::min` and `std::max`; fixes needed.

On Windows `min` and `max` are also macros defined by `Windows.h` (when included without defining `NOMINMAX`), so `std::min` and `std::max` will fail to compile. We can either arrange to define `NOMINMAX` in `DemoCamera.vcxproj` or use the `(std::min)(a, b)` workaround. Probably the former.

marktsuchida:
Set `NOMINMAX` at the project level for `MMCore` and `DemoCamera`. It would be nice to do this for all projects, but there are about 10 that break (i.e., currently depend on the Windows.h macros), so that will be another project.

PR #413, opened 2023-12-11
Remove printf statements during HID device init

### LINKED ISSUE #402: HIDManager should not print to stdout (by default, at least)
See https://github.com/pymmcore-plus/pymmcore-plus/issues/286.

Device adapters should avoid printing to stdout or stderr, since people might use them from the command line (also, nobody will see the stdout/stderr when running in MMStudio, at least on Windows using the regular launcher). In this particular case, we could probably just remove the logging.

(Actually a better solution would be for MMDevice to provide a way to write to the CoreLog before devices are initialized, but that's a separate story.)

tlambert03:
closes #402

@nicost, i believe you wrote the HID manager ... can you review this one and confirm whether it would be OK to remove these print statements?


marktsuchida:
To me this looks good; we can re-add the logging when/if I get around to supporting logging prior to device initialization. But I don't know the ins and outs of HID device enumeration, so will wait for @nicost's thoughts.

Issue #243, opened 2022-09-21
More feedback and prototyping of new camera API (chapter 3)


henrypinkard:
This issue is for providing feedback on the [new camera API](https://github.com/micro-manager/mmCoreAndDevices/blob/main/camera_triggering_API_v2.md). 

With this v2, we're now in a prototyping phase where I will be implementing it on a Basler camera.

Previous discussions of this API can be found in:

[Chapter 1](https://github.com/micro-manager/mmCoreAndDevices/issues/84)

[Chapter 2](https://github.com/micro-manager/mmCoreAndDevices/issues/226)

Version 1 of the proposed API can be found [here](https://github.com/micro-manager/mmCoreAndDevices/blob/4ce6d521a3875c0ecdd691fba5751b6a43bb0cd1/camera_triggering_API.md)

marktsuchida:
@henrypinkard Sorry for the silence - I'll try to comment on this and your other proposals in the next few days.

henrypinkard:
@marktsuchida No worries, I am mid-prototyping on this one right now and am accumulating questions as a go. I think it would be more efficient to batch it all into one larger discussion later

henrypinkard:
*Accumulating questions and fixing mistakes

marktsuchida:
@henrypinkard Regarding the timing events (such as `CameraEventFrameEnd`),
- These seem like a good opportunity for breaking up the problem into smaller units: presumably these can be added in a second round, completely independent from the main trigger configuration API.
- Can you point me to one or more actual camera API docs that correspond to these events? I'm trying to think about whether they should be callbacks more strongly associated with individual acquisitions rather than being "events".

henrypinkard:
> * Can you point me to one or more actual camera API docs that correspond to these events? I'm trying to think about whether they should be callbacks more strongly associated with individual acquisitions rather than being "events".

https://docs.baslerweb.com/event-notification

and from GenICam

![events](https://user-images.githubusercontent.com/7969470/196295851-340c9295-de95-416b-8559-7715245672da.PNG)

![lineevents](https://user-images.githubusercontent.com/7969470/196297180-22267825-a541-41f1-aeb7-f167450b345e.PNG)


So far as I can tell, events/callbacks like ExposureEnd/Line0RisingEdge/etc (or some equivalent) are needed to implement triggers and acquisitions correctly in the GenICam way of doing things.

[...]

PR #414, opened 2023-12-11
PyDevice PR

JeroenDoornbos:
PyDevice is an adapter that imports objects from a Python script and integrates them into MicroManager as devices (e.g. a camera or a stage). This integration enables the use of Python scripts to control microscope hardware, without requiring any programming experience from the end user. Essentially, PyDevice acts as a translator, allowing Python-developed objects to be used in MicroManager with almost no interfacing code required.

We have spent a while working with it and integrating it with our OpenWFS framework. The idea is that both developers and users can have the control they require, without sacrificing modularity and development speed. The OpenWFS repo: https://github.com/IvoVellekoop/openwfs also contains some more advanced devices, such as a laser-scanning object that functions as a MicroManager camera. 

We recommend the README.md or example_devices/PyDevice.md as starting points.

It was built by Ivo Vellekoop and Jeroen Doornbos, please let us know what you think! We are very interested in your insights.


marktsuchida:
Thanks for the PR! I've been wanting Python-based device adapters for several years, so I'm excited to review this in detail -- there's lots going on here!

A few quick initial thoughts:
- It looks like this starts a Python interpreter inside the Micro-Manager device adapter. Since only one CPython (non-sub) interpreter can exist per process, presumably the Python interpreter is initialized (`Py_InitializeEx()`) when the PyDevice hub is first initialized, but not on subsequent initializations. Can we be confident that state is not carried over from earlier initialization (or does that matter?)? And perhaps more importantly, what happens if this is loaded from pymmcore?
    - If the latter doesn't immediately work, we can probably come up with some kind of safeguard so that PyDevice will safely refuse to load under pymmcore. But maybe it just works? I'll test it when I have a chance but let me know if you already know the answer.
- I saw a few places where you supply `std::monostate` as the template argument to the DeviceBase classes. Intuitively, I would expect this to not work at all, but clearly it must work(?) -- can you explain what is going on here?
- Your interpretation of the `Home()` methods of stages is incorrect: it should perform the hardware homing operation of the stage (for applicable stages). It should not go to position 0 (or (0, 0)). I haven't reviewed all the device actions yet, but this one caught my eye.
- Do you intend the OpenWFS repository to be where Python devices will be collected? Do you have any thoughts on how to support API evolution in the future?

tlambert03:
yeah, wow... this looks quite heroic ❤️.  I also look forward to digging in to see what you've done here!

> And perhaps more importantly, what happens if this is loaded from pymmcore?

1. yeah, this is probably my first question as well after a quick peek at the code.  Will this approach be permanently limited to running its own interpreter?  or is it conceivable that `PyObj::InitializeInterpreter` could somehow use an existing python interpreter if, for example, we're already controlling the C-level micromanager code using pymmcore[-plus]?    Perhaps a related question would be: do you think that's even a valuable use case?  or was the goal here specifically just to enable easy python-based device control within the Java micro-manager application?

> allowing Python-developed objects to be used in MicroManager with almost no interfacing code required.

2. on first glance, it looks like `bootstrap.py` might require a good amount of discussion.  It *is* defining a new interface after all that various device types have to follow.  (yes, it removes a ton of boilerplate, which is really quite nice! but it does establish new patterns/APIs that python users going forward must adhere to)

3. it would be great if astropy wasn't a hard dependency.  I gather you used it for a more easily inspectable runtime object (for the purpose of using `@runtime_checkable`) but perhaps a custom approach could use something like `typing.Annotated[float, 'micron']` (i.e. something that comes with the standard library), or maybe `pint` (which has no additional dependencies).

anyway, clearly a tremendous amount of work here.  congrats.

IvoVellekoop:
Dear Mark, 

Thank you for your detailed review. Let me answer your questions:

- _It looks like this starts a Python interpreter inside the Micro-Manager device adapter. Since only one CPython (non-sub) interpreter can exist per process, presumably the Python interpreter is initialized (Py_InitializeEx()) when the PyDevice hub is first initialized, but not on subsequent initializations. Can we be confident that state is not carried over from earlier initialization (or does that matter?)?_

Indeed, the interpreter is initialized only once, and never de-initialized. Unfortunately, this is the only option. Initially, we called Py_FinalizeEx() after the last PyDevice was closed. However, according to the 'bugs' section of the Python API documentation (https://docs.python.org/3/c-api/init.html#c.Py_FinalizeEx)  "Some extensions may not work properly if their initialization routine is called more than once; this can happen if an application calls Py_Initialize and Py_FinalizeEx() more than once.". Especially, this caused serious problems with the numpy package, and in the end we chose to always keep the interperter open.

As a result, a script may change some global state which could be visible to some other script that is loaded later. I do not think this is really a problem in practice. We do reset the sys.path variable between initializations, so that plugins cannot affect each other through the module search path.

_-  And perhaps more importantly, what happens if this is loaded from pymmcore? If the latter doesn't immediately work, we can probably come up with some kind of safeguard so that PyDevice will safely refuse to load under pymmcore. But maybe it just works? I'll test it when I have a chance but let me know if you already know the answer._

We did some preliminary tests combining PycroManager and PyDevice and all worked as expected. If I understand correctly, PycroManager runs in a separate process, so it can safely use its own Python interpreter.

_- I saw a few places where you supply std::monostate as the template argument to the DeviceBase classes. Intuitively, I would expect this to not work at all, but clearly it must work(?) -- can you explain what is going on here?_

In fact, the CDeviceBase class does not really use the template argument. Since the 'curiously recurring template' pattern complicates the code a lot, we are just using a dummy class as template argument. Ideally, I think the template argument should be removed from CDeviceBase completely. 

To be precise, if you look at the source code of CDeviceBase, the template argument is only used for two typedefs (CPropertyAction and CPropertyActionEx), and these typedefs are only used in CreatePropertyWithHandler, which wraps the handler in a CPropertyAction, which is a specialization of CAction, which is a subclass of ActionFunctor. All other CreateProperty functions just take an ActionFunctor as input, which is automatically type safe and does not need a class template argument. So, except for the somewhat redundant CreatePropertyWithHandler function, the template argument is not used.

_- Your interpretation of the Home() methods of stages is incorrect: it should perform the hardware homing operation of the stage (for applicable stages). It should not go to position 0 (or (0, 0)). I haven't reviewed all the device actions yet, but this one caught my eye._

Thank you for spotting this. I was a bit confused about what should happen. My interpretation was the following:
- getting the stage position: returns the current position, even when the stage is moving
- setting an absolute position: set's the intended target position, does not wait for the stage to arrive at that point ('busy()' can be used to wait for the stage to reach the position)
- setting a relative position: set's the intended target position relative to the previously set position (not relative to the current position, since the stage may still be moving)
- homing: performs hardware homing of the stage, does not wait for the homing to complete. Set the target position of the stage to 0 (since this where the stage will be after homing completes).

However, I understand this last part (or perhaps more) is incorrect. Can you clarify what is the intended behavior so that we can fix it? 
As for the OnPositionStaged callback, when should it be called? when the movement starts, or after it completes? (which will be a lot harder to implement).

_- Do you intend the OpenWFS repository to be where Python devices will be collected? Do you have any thoughts on how to support API evolution in the future?_
At this point, we decided to separate the MicroManager adapter and the OpenWFS repository since the two pieces of software can be used completely independently. Still, OpenWFS provides some extremely useful functions, such as the ability to completely simulate a motorized microscope, including aberrations and wavefront shaping. We intend to actively maintain OpenWFS as the main repository for our (stable) research code, and encourage contributions from the community.

If the MicroManager API changes, we will be happy to update PyDevice. There is some work to be done supporting other devices like filter wheels. Even though those devices are not the highest priority for us at the moment, we are happy to help implementing a binding for those device types if someone wants to use it.

tlambert03:
> > - And perhaps more importantly, what happens if this is loaded from pymmcore? 

> We did some preliminary tests combining PycroManager and PyDevice and all worked as expected. If I understand correctly, PycroManager runs in a separate process, so it can safely use its own Python interpreter.

pycro-manager and pymmcore are two separate things.  As you likely know, pycro-manager controls the core via RPC to another running java process.  So, this PyDevice approach works fine in that setup, since python is already a secondary player of sorts.  Pymmcore (and pymmcore-plus on top of that), by contrast, directly controls the core in the same process via C-extensions.  I can't speak for the core developers of micro-manager, but I know many of us would love to see a future in which Java is not *required* to accomplish meaningful things in micro-manager.  In that model, it would be much nicer to have an API design that allows a future in which python devices can be controlled in the same process.  

Can you comment specifically on your thoughts about the possibilities of running this in the same process as the core? (that is *not* using pycro-manager)?

IvoVellekoop:
Dear Talley,

Thanks for the compliments!
And I understand the difference now, I have already edited my response.

Just to add to this, our core design philosophy was to have code that works in Python (where we do our algorithm development, simulation, testing, and many of the experiments), and to make that code available to the end user (who will use the MicroManager GUI) without the need to change a single line of code. So, all our device adapters run both with and without the C++/Java detour.

About your other remarks:

_-on first glance, it looks like bootstrap.py might require a good amount of discussion. It is defining a new interface after all that various device types have to follow. (yes, it removes a ton of boilerplate, which is really quite nice! but it does establish new patterns/APIs that python users going forward must adhere to)_

I agree, it indeed should be reviewed. We could also perhaps make some simple wrapper to make the interface compatible with pymmcore.

_- it would be great if astropy wasn't a hard dependency. I gather you used it for a more easily inspectable runtime object (for the purpose of using @runtime_checkable) but perhaps a custom approach could use something like typing.Annotated[float, 'micron'] (i.e. something that comes with the standard library), or maybe pint (which has no additional dependencies)._

I agree that the dependency is not so nice. However, we do use it extensively in our code and it caught a lot of bugs. At some point we lost a lot of time because a student specified a wavelength in micrometers instead of nanometers, and now such problems are caught at runtime or avoided since units are converted automatically (which does not happen with the string annotations). I'll have a look at pint to see if it can replace astropy.units since indeed it is quite a heavy dependency if you don't use Anaconda.

marktsuchida:
@IvoVellekoop Thank you for your detailed (and mostly satisfying) explanations!

> We haven't tried pymmcore integration yet. I am not sure about the use case, since why would you use MicroManager to load a Python script into a Python environment?

I think you are vastly underestimating the potential of what you have created :)

With the ability to write device adapters in Python, we can imagine having an official collection of such adapters, and bundling them with the usual Micro-Manager installer. They could "just work" in the Hardware Configuration Wizard and the user doesn't (necessarily) need to know that they are implemented in a different language (beyond perhaps having to install Python and a few things). In such a world, it would be unfortunate to artificially exclude users of pymmcore-based apps from using these devices. (There are additional reasons why you might want to go through MMCore for all devices, too, such as uniformity of the interface and clear separation of concerns.)

Which brings me to the main point I want to address before merging this (I'll talk about the more detailed things in separate posts later).

Once PyDevice is merged (or _if_, but I hope we'll merge this product of your hard work!), it will become the only (and therefore likely the de facto "official") way to write device adapters in Python -- especially given its generic-sounding name. This means that it needs to cater to a wide audience -- basically almost all users of mmCoreAndDevices -- because it will become harder to propose an alternative mechanism for having Python devices (and it would be unfortunate to duplicate the effort anyway). This does not mean that it needs to be everything to everybody from the get go; only that, IMHO, we need to have a clear picture of how we will be able to evolve PyDevice into something versatile and robust and maintainable for a long time. (And yes, this is perhaps a slightly higher standard than MMDevice's design might suggest, because I'd hate to repeat past mistakes.)

> If the MicroManager API changes, we will be happy to update PyDevice. There is some work to be done supporting other devices like filter wheels. Even though those devices are not the highest priority for us at the moment, we are happy to help implementing a binding for those device types if someone wants to use it.

I appreciate your commitment here, and in any case anybody who proposes changes to MMDevice will need to make the corresponding changes to PyDevice if necessary. I'm also not worried about adding the remaining device types -- that can be done as needed over time. What I do think we need to solve is how to deal with backward-incompatible changes in MMDevice and PyDevice.

Changes to MMDevice may include not just adding methods (which usually will have default implementations and thus be easy to deal with) but also removing methods (including methods for devices to call into MMCore) or changing method signatures (we allow such changes as long as they either maintain existing behavior or fail to compile the old usage -- provided that there is a realistic migration path that doesn't involve testing all devices). Since we have all C++ device adapters under the micro-manager GitHub org (and mostly in this repo, although this may change in the future), we can occasionally make sweeping changes that improve MMDevice. Any out-of-org adapters can be updated at the maintainer's responsibility (though not a recommended way to do things). (In addition to such source-level considerations, we try to ensure that the ABI between device adapters and MMCore is stable, incrementing the "device interface version" (currently 71) when this is not the case. MMCore rejects to load incompatible device adapters.)

We will probably also want to be able to evolve the Python interface requirements of PyDevice even when MMDevice is not changing (some of the points that @tlambert03 is raising being potential examples of the kinds of such changes).

Having the Python devices declare a version number might be part of the solution, though I'm not entirely sure if that is the best or even a desirable solution -- it could be quite error-prone unlike the automated ABI version of MMDevice. Or maybe PyDevice can somehow be made to work with different versions of the interface (or detect and reject outdated ones). I don't yet have a clear suggestion for how we should do this (will keep thinking and studying your code), but I think it is important to design for change; if we get that part right, we are much more likely to be able to deal with any other issues as they arise.

One thought/available option: Since users are already required to have numpy and astropy (although, in theory, both of these dependencies could be dropped), it wouldn't be hard to require another pip/conda-installable dependency that is specific to PyDevice (potentially replaceing bootstrap.py). So compatibility checks/bridging might be something that can be solved mostly on the Python side, keeping the interface between (the package replacing) bootstrap.py and mmgr_dal_PyDevice more rigid and strictly versioned.

tlambert03:
> > We haven't tried pymmcore integration yet. I am not sure about the use case, since why would you use MicroManager to load a Python script into a Python environment?

> I think you are vastly underestimating the potential of what you have created :)
> With the ability to write device adapters in Python, we can imagine having an official collection of such adapters, and bundling them with the usual Micro-Manager installer. They could "just work" in the Hardware Configuration Wizard and the user doesn't (necessarily) need to know that they are implemented in a different language (beyond perhaps having to install Python and a few things)

this! 👆 ❤️

There have been many discussions (in and outside of this repo) around making it easier for people to create device adapters written in python, and use them along-side device adapters written in C++.  At least three alternative packages to micro-manager have been created with the primary motivating factor being to achieve something like this (making it easier to write device adapters in python).  As such, I think it is _very_ important to usher `mmCoreAndDevices`, with its awesome library of pre-existing device adapters, into a future model that allows devs to contribute pure-python device adapters and use them side by side the existing libraries.  I would be concerned about the longevity of an approach that kind of bakes in the concept of "micro-manager as a java gui that controls a c++ library, with python as an optional secondary language to simplify device adapter creation".

And, as @marktsuchida points out... as soon as a solution that appears to facilitate pure-python device adapters gets merged in mmCoreAndDevices, it does indeed become the de-facto approach that needs to be discussed – along with it's subtleties and use-case assumptions – each time a new user comes along and says "I'd like to write a device adapter for micro-manager in python".  So, at the very least, it's going to be important to consider the naming, messaging, READMEs and optics around a device adapter like this.

IvoVellekoop:
@tlambert03 and @marktsuchida. Ok, integration with pymmcore indeed makes a lot of sense. We'll try to get it working in the next couple of days. However, I'll need some time to test it properly, especially the multi-threading and possible interference with the running interpreter (such as setting sys.path). 

>  we need to have a clear picture of how we will be able to evolve PyDevice into something versatile and robust and maintainable for a long time

The way I see it, for the compatibility, there are two main questions:
* How to deal with changes in the interface on the micromanager side?
* How to deal with changes in the interface on the Python side?

The current design is loosely coupled, in the sense that it is possible to accommodate changes in the mm device interface without requiring a change in the Python code, and vice versa. This means that we can treat the issues above separately to a large extent.

**On the micromanager side**
I suggest absorbing changes in the mm device interface into the C++ code as much as possible. If functionality is removed, PyDevice C++ code should use alternative functions or (worst case) replicate the removed functionality. If functionality is added, the Python interface may need to be expanded with optional properties or methods (see below).

Since changes in the device interface version number require rebuilding the plugin, possibly after modifying some of the source code, there is no backward or forward compatibility on the mm side. This is a good thing because the C++ code only needs to consider a single version of the device interface. To facilitate these changes, it makes sense to keep the code in the micromanager GitHub site as @marktsuchida mentioned.

**On the Python side**
I think we should start with the simplest interface that covers the basic needs and indeed review it properly. 

In future releases, we can expand the interface with optional functionality (for example, adding a wait() function instead of just polling busy). 'bootstrap.py' can then be updated to provide default implementations to maintain compatibility with older Python code. We are already doing this to some extent, by adding a 'binning' property to a camera object if it is missing from the Python object.

About versioning, I think duck typing is the 'Pythonic' way to handle such things, instead of using version numbers or predefined base classes. By having the Python code in bootstrap.py inject default implementations for missing functionality, we should be able to ensure forward and backward compatibility unless there are really deep changes. In that (unlikely?) case, we could define a new interface altogether, and just let bootstrap construct a wrapper for objects using the old interface. This will work as long as Python can distinguish the different object classes by run-time inspection.

> One thought/available option: Since users are already required to have numpy and astropy (although, in theory, both of these dependencies could be dropped), it wouldn't be hard to require another pip/conda-installable dependency that is specific to PyDevice (potentially replaceing bootstrap.py). 

We did think about having such a package. For the moment, however, it is not needed since no PyDevice-related code is needed on the Python side. This allows users to write code in vanilla Anaconda (which includes numpy and astropy) and it allows the Python object to be fully functional without any micromanager related libraries.

Still, is will be useful to have a package with PyDevice-compatible devices and utility classes. We put all such classes (such as the scanning microscope 'camera' and the virtual microscope) in the OpenWFS package.
If we can get the pymmcore integration to work, it may make sense to put some utility functions etc. in the pymmcore package later.

Currently, bootstrap.py is hardcoded in the C++ code. The advantage is that bootstrap.py only needs to work for a single mm device interface version and a single version of the code. Since bootstrap.py also plays a role in absorbing changes to the mm device interface, it is strongly coupled to the C++ part rather than to the Python part. The disadvantage is that it cannot be updated without recompiling the code, so any update will have to go hand in hand with a micromanager update. Perhaps this disadvantage can be solved later by indeed including part of the bootstrapping code in the pymmcore package.

Finally, currently we do need numpy for transferring data from camera frames. The astropy.units are used to make sure that users don't accidentally mix e.g. steps and micrometers for stage positions, or microseconds or seconds for the camera shutter time, etc. We all remember the Mars Orbiter :-)
I think it will be relatively easy in the future to extend the plugin to use alternative unit packages (e.g. pint) or other types of arrays (perhaps pytorch?).

But for now, how to proceed? How about the following:

1. Review the Python device interfaces in bootstrap.py - github discussion
2. Review the rest of PyDevice - github discussion
3. Fix any issues coming up in the review (such as stage homing) - University of Twente team
4. Make PyDevice compatible with with pymmcore - University of Twente team, separate pull request
5. Include in the nightly build
6. Monitor user feedback and fix possible issues before including in the stable release.

marktsuchida:
Thank you, @IvoVellekoop, for taking the time to explain your thoughts. I think these directions are great. The most important thing to me is that we generally agree to the goal of (ultimately) making this into something that can be supported long-term, and I appreciate your willingness to work with us with that in mind!

- I like your proposed steps (1-6). Let's get started with discussing the details of bootstrap.py (I'll try to have my own comments in the next ~day).
- I agree that pymmcore compatibility can be a separate (later) PR. Let's just make sure that merging this PR will not crash pymmcore when accidentally loading PyDevice. (Perhaps detect when Python has been initialized before PyDevice ever initialized Python.)
- If we decide to replace bootstrap.py (or part of it) with an independent Python package, those items probably do not belong in pymmcore but rather a separate package ("mmpydevice"?), because you don't need pymmcore to have Python devices, and vice versa (they are two distinct APIs). In any case, that is probably something to be done in a future PR.
- If we eventually decide to ship Python device adapters with Micro-Manager, we might want to host the code for such devices here. (Of course, we'd also want to keep it easy to use external ones, too.) But that is many steps away, so let's not worry about it now.

henrypinkard:
Those next steps sound good to me

Great work @JeroenDoornbos  and @IvoVellekoop! Really excited about this and the new possibilities it will open!

henrypinkard:
Since Pycro-manager now supports both pymmcore and the MMCoreJ as backends it could be useful (and I'd be happy test) if that these work the same from the user perspective

marktsuchida:
I've looked a little more carefully at the camera device and noticed that it uses the problematic "default" implementation of sequence acquisition that currently resides in `CCameraBase`. (Not your fault for not knowing that this is not recommended practice. I've now added #421 that explains this.) We'd want to fix this but I'd be okay with doing so after getting this PR merged.

The core devs (@nicost, @henrypinkard, and myself) discussed how to proceed with this during a meeting today. How does everybody (especially @IvoVellekoop, @JeroenDoornbos, and @tlambert03) feel about merging this sooner rather than later, but making it clear in the documentation that the interface used by the Python devices is experimental (under development) for now and subject to incompatible changes, until we're all happy with it (with the rough idea that this may take several months)? Proposing this mostly because it will probably be easier to discuss and implement individual changes (or overarching ones, for that matter) as separate issues and PRs.

IvoVellekoop:
@marktsuchida. Thanks for the detailed comments. I am perfectly fine with merging a 'beta' version first and working on the remaining issues (and the ones that may come up later) in separate PRs.

@henrypinkard. I've worked on getting PyDevice to work in pymmcore, but I currently experience an issue with initializing the numpy C API. Initializing the numpy api is quite fragile and error-prone (see, for example [one of many posts on SO](https://stackoverflow.com/questions/59314230/python-c-api-crashes-on-import-numpy-when-initilizing-multiple-times)).
I think it is possible to remove the numpy dependency altogether by using the [Buffer protocol](https://docs.python.org/3/c-api/buffer.html). This has the huge advantage that the Python program then can also use other data sources, such as PIL and PyTorch. Moreover, it eliminates the numpy dependency. At a later stage, we could also remove the strong astropy.units dependency so that we only need the core Python libraries.
The only drawback is that we need at least Python 3.11 for this to work, but I think that is a fair requirement, since Python 3.10 is already out of active support anyway. 
I'll work on this the coming time.

tlambert03:
> How does everybody (especially @IvoVellekoop, @JeroenDoornbos, and @tlambert03) feel about merging this sooner rather than later, but making it clear in the documentation that the interface used by the Python devices is experimental (under development) for now and subject to incompatible changes, until we're all happy with it (with the rough idea that this may take several months)?

I can see the merit of getting something into main so it can be experimented with, though I do think it could use a lot more discussion and refinement.  It's not ultimately up to me of course, but things I'd encourage (mostly around compatibility):
- do expose buffers using the standard library buffer protocol
- i think requiring py 3.11 is actually a bit strict. are you saying that because `Py_buffer` is part of the stable ABI as of python 3.11?  If so, am I correct in assuming that you could still support earlier versions (since `Py_buffer` itself was around well before 3.11) provided you only used parts of the structure that we common to the least version you want to support? I would propose to support all versions which are not end-of-life, which [currently goes back to python 3.8](https://devguide.python.org/versions/)
- i'd put removing astropy very high on the list of priorities (perhaps even before merging, but not my call).  I understand the goal of preventing units errors, but I think there are many ways to do that without requiring astropy.  You could have cascading options for checking the return values
    1. check if `'astropy'` is in `sys.modules`, and only then check `isinstance`
    1.  check if `'pint'` is in `sys.modules`... (i.e. let people use pint instead, which is a lighter requirement)
    1. allow people not to use either, but support `Annotated[float, 'micron']` or something like that 
    1. let people make their own decisions as to whether they want to opt in to this, and simply allow returning a float
- i might actually reconsider the use of `runtime_checkable`. It *does* make for a nice looking interface, particularly for public facing `Protocols` that you want people to implement *and* be able to import.  But in this case, it doesn't really look like you're encouraging anyone to actually import those classes in `bootstrap.py` right? (which is a good idea), and  `runtime_checkable` is a strictly internal convenience, right?  It really is a rather limited function.  I'll note the following two things from the [docs](https://docs.python.org/3/library/typing.html#typing.runtime_checkable):

    - > runtime_checkable() will check only the presence of the required methods or attributes, not their type signatures or types.
    - > An [isinstance()](https://docs.python.org/3/library/functions.html#isinstance) check against a runtime-checkable protocol can be surprisingly slow compared to an isinstance() check against a non-protocol class. Consider using alternative idioms such as [hasattr()](https://docs.python.org/3/library/functions.html#hasattr) calls for structural checks in performance-sensitive code.
    
    just pointing that out in case part of the attachment to `astropy` was an assumption that it was helping you with type checking.  If not, i'd say just let the users use astropy, or pint, `Annotated`, or even just `float` if they want to, and support all of them.

- lastly, if you can stomach it, I'd suggest perhaps putting the experimental nature of this directly in the device adapter name until interested parties have had time to play with it and weigh in.  Like `ExperimentalPyDevice` (or similar).  I do think the very biggest hurdle to cross in establishing the protocols that we'll all be required to support going forward is simply getting the namespace into main :) so, it's not trivial to step many of these things back in the future. 

that said, thanks again for all the great work @IvoVellekoop and @JeroenDoornbos.  It definitely is an exciting approach and i look forward to seeing how it evolves.

IvoVellekoop:
@tlambert03 Thank you for sharing your thoughts. I have some thoughts on addressing these issues, and it would be good to have your thoughts (including those of @marktsuchida and @henrypinkard) before implementing everything. In order of priority:

0. **Non-intrusiveness**. A leading design principle for PyDevice was to have code that runs in a regular Python program be imported as a device without any modification. This means that we do not require the Python objects to have a certain base class, PyDevice-specific annotations, or other stuff that the developer does not already need on the Python side. I realize that the `astropy.units` dependency violates this principle, so it should indeed be removed (see below).

1. **Compatibility**. Ideally, we would like users to experiment with PyDevice now, and keep those .cfg files valid in the future, even if the python API changes. I suggest achieving this with a `VersionNumber` property on the `PyHub` in micromanager. This way we don't need different device names for the current (experimental) and future (stable but evolving) versions. The `bootstrap.py` script is passed the version number as a parameter and constructs an appropriate wrapper to convert old version api objects to the latest version. I think this will just be a few dozens of lines of Python code. For simplicity, we don't allow mixing objects of different api types in a single hardware configuration.

2. **`astropy.units dependency`**. I propose modifying the python api to have `_um` and `_ms` suffixes for the properties that have these units (such as the step size of a stage). This seems like the least intrusive and simplest option. To be able to work with `astropy.units` and `pint`, the bootstrapping script is modified to recognize properties with a signature like `step_size -> Quantity[u.um]` and automatically adds the corresponding `step_size_um` version of the properties through monkey patching, creating a wrapper, or another mechanism. This has the added advantage that all unit conversion is done in Python, which simplifies the C++ part quite a bit. 

3. **`pymmcore`** interoperability. I am working on this, but there are still some problems with initialization/deinitialization. I am not sure anymore if these are `numpy` related. More on this subject later.

4. **introspection and testing**. I agree that `runtime_checkable` has limited use because it does not check the signatures of the functions and properties. However, Python functions really don't have a signature, only an optional type hint which is not enforced. The 'Pythonic' way is to not check types at all, but just have the runtime raise an `AttributeError` or `NameError` at some point if an object does not support some action. For embedding into micro-manager this makes it very hard to develop and debug device adapters as type errors can only be detected at run time. What would potentially be very useful is a Python module that developers can use to test the devices they created. This module would detect the object type just as `bootstrap.py` does, and have some testing facilities for reading and writing all properties (verifying if the data types are correct at runtime), grabbing a frame, etc. It could also give diagnostic information about what properties/methods are missing for an object to be recognized as a camera, stage, etc. This module would not be a hard dependency since it is only used for testing. We could put this code in our `openwfs` package, which we are releasing in January and which already includes some examples (such as a fully simulated microscope). I am also open to putting it in `pymmcore` or `pycromanager`.

5. **`numpy` dependency**. If we can get `pymmcore` to work without dropping the `numpy` dependency, I don't think moving to the buffer protocol is very urgent. The thing with the Python versions is really problematic. AFAIK, using the limited API is the only way to use the currently installed version of Python. With the limited API, the device plugin is linked with import library `python3.lib`, which locates `python3.dll` on the host system at run time. `python3.dll` is just a stub that forwards to `python39.dll`, or whatever version is installed on the user's system. If we want to use functions that are in `python39.dll`, but not in `python3.dll`, we would need to link to `python39.lib` explicitly. This is how we set it up before, but then it doesn't work if the user has a different Python version installed. Only with Python 3.11 the buffer protocol is part of the stable api (`python3.dll`), but indeed it is a bit too early to expect everyone to have this version installed. Therefore, I suggest trying to get `pymmcore` to work with numpy and park this for now. 

We'll work on these points after the holidays. As always, all feedback is appreciated
Enjoy the holidays!

tlambert03:
> This means that we do not require the Python objects to have a certain base class

To be clear, I like the protocol 
/interface pattern a lot. I'm not proposing a mandatory base class. I'm just suggesting that the runtime_checkable method that you're using might not be as robust as you'd like (if you were using it so that you could use `isinstance`)

> To be able to work with astropy.units and pint, the bootstrapping script is modified to recognize properties with a signature like step_size -> Quantity[u.um] and automatically adds the corresponding step_size_um version of the properties through monkey patching, creating a wrapper, or another mechanism

Yep I also recognize how you're doing this.  But think it would be possible to support all these things without a top level import (and therefore dependency) on astropy. I think you could support astropy and pint (and additional methods if you want) without putting import astropy at the top of bootstrap.py

henrypinkard:
I'll try to take a deeper look at this and give some feedback some time next week

henrypinkard:
@JeroenDoornbos and @IvoVellekoop, I'm just looking through this, and I will have more detailed feedback soon. But at the moment I'm hung up on an error: when I try to add a device in the hardware configuration wizard (like the `example_xystage.py`), the following error pops up:

error1

Any thoughts on what might be causing this and how to fix it?

marktsuchida:
Okay, lots to catch up here. Thanks for your comments @tlambert03, @IvoVellekoop, and @henrypinkard.

I generally agree that dependency on numpy and astropy should be avoided especially in the hard-coded C++ part of this. But this can be addressed after merging.

@tlambert03 wrote:

> * lastly, if you can stomach it, I'd suggest perhaps putting the experimental nature of this directly in the device adapter name until interested parties have had time to play with it and weigh in.  Like `ExperimentalPyDevice` (or similar).  I do think the very biggest hurdle to cross in establishing the protocols that we'll all be required to support going forward is simply getting the namespace into main :) so, it's not trivial to step many of these things back in the future.

I agree with this in spirit. But I think it is good enough to have a device page (micro-manager.org/PyDevice) that starts with a prominent warning that this is experimental and both Python code and .cfg files will break without notice ("use at your own risk"). The device description (second arg to `RegisterDevice()`) should also say "experimental".

@IvoVellekoop wrote:

> 1. Ideally, we would like users to experiment with PyDevice now, and keep those .cfg files valid in the future, even if the python API changes.

This is a good goal to uphold once PyDevice is no longer "experimental". I do not think the design is at a stage where we can even suggest that this is likely. (If we get lucky and they do stay valid, that's great of course.) Even if the device is not named "experimental", we should make it prominently clear to users that things will break. (More on the versioning idea below.)

I support merging this early primarily because I think it is more productive if those of us who want to improve the design can propose improvements in the form of issues and PRs (hopefully smaller PRs that target a single issue at a time). I do not support merging if we're going to claim that anything about PyDevice will keep compatibility into the future at this time.

@henrypinkard wrote:

> From my perspective, things that would need to happen before merging are:
> 
> * [ ]  Adding python to 3rd party public
> * [ ]  Document that this is experimental and add versioning
> * [ ]  Fix the merge conflict in micromanager.sln

As I think I already mentioned somewhere above, I can take care of the Python/NumPy dependency in the build (in 3rdpartypublic; directory names will have to change and I'll update the project file). And I can fix the merge conflict in the .sln. There are other administrative things (for example, .gitignore shouldn't suddenly blanket-exclude .dlls at this time) that I'll also fix outside of the PyDevice directory. These are standard steps I take when merging a new device adapter and are best done just before merging (by pushing to the PR branch).

I suggest that we do **not** add versioning at this time. That is a complex enough topic that I think should be discussed in a separate issue or PR, and in any case we do not want to appear to promise backward compat. Also it takes a lot of extra work to keep backward compatibility, and I think it is premature to burden ourselves with that complexity (not to mention that it would prevent us from removing dependencies).

(I also am not sure I understand the utility of adding a version number for the Python interface (if I got that right) in the .cfg file. It is the devices written in Python that know what interface version they were written to, not the end user configuring devices. In general .cfg files should only break if devices remove or rename (or change allowed values of) a property; this is usually orthogonal to what version of the Python interface is being used under the hood.)

I'm happy to create the device page (with the experimental warning) and link to the Markdown files in the source tree. I'll Cc all of you in the PR. The page can be expanded, especially after we're out of the experimental phase.

The only other thing I was worried about was poisoning pymmcore users (including on Python 3.8-3.10) who enumerate all available devices, but presumably this is not an issue because no Python C API is called until `CPyHub::Initialize()`. It would be even better if we ensure that `Initialize()` doesn't crash under those conditions, but that can be a separate PR if necessary. (Or maybe you'll have actual support for using with pymmcore soon enough.)

So as far as I'm concerned, unless there are any objections, I'm ready to fix up the build files and merge. I'll wait until I hear back or Thursday before proceeding.

henrypinkard:
> I suggest that we do **not** add versioning at this time. That is a complex enough topic that I think should be discussed in a separate issue or PR, and in any case we do not want to appear to promise backward compat. Also it takes a lot of extra work to keep backward compatibility, and I think it is premature to burden ourselves with that complexity (not to mention that it would prevent us from removing dependencies).

Yes, I see your logic on this. So I think its reasonable to push this off to a future PR if okay with @IvoVellekoop  and @JeroenDoornbos. It will also be nice to have future improvements in smaller, more digestible PRs to make it easier to stay on top of things.

I'm happy to (once it's merged) make a post advertising this on image.sc announcing it and asking for more testing/feedback, unless @IvoVellekoop or @JeroenDoornbos  you would prefer to do so?

marktsuchida:
> I'm happy to (once it's merged) make a post advertising this on image.sc announcing it and asking for more testing/feedback

Not that anybody can forbid you from doing this :), but I do feel that it's a bit early for this. I'm certainly against making it sound like an official announcement from the Micro-Manager developers.

I think people who are closely watching MM development will find this PR. For people who are more scientists than programmers and would like to control devices with Python, but are not interested in how precisely the underlying mechanism works (i.e., a big part of the target audience of this feature), it will just be frustrating if we proceed to repeatedly break the Python interface (which I believe we will).

I think we will get better user feedback if we ask for it once we have settled many of the issues mentioned above (and possibly more that some of us (at least me) haven't yet mentioned).

Note that many of the issues are about long-term usability and maintainability (not just of the PyDevice adapter and the Python modules written to work with it, but also MMDevice/MMCore which need to continue to support PyDevice). These software engineering issues are not things that, in my experience, we are likely to get productive feedback on from casual testers, so it would be prudent to avoid giving the impression that PyDevice is "ready for testing" (I don't think it really is).

I wouldn't be opposed to a more open-ended post saying that we are _starting_ to develop device control in Python (because that is the stage I believe we are at), and that we would like to hear from potential users about their potential use cases and requirements.

And, by the way, I wouldn't have anything against the contributing authors (@IvoVellekoop and @JeroenDoornbos) making an announcement from their own perspective.

henrypinkard:
@IvoVellekoop @JeroenDoornbos do Mark's comments above look good to you all? Ready for this to be merged?

IvoVellekoop:
Hi all (@marktsuchida, @tlambert03, @henrypinkard, @JeroenDoornbos),

It has been a while, but I finally found the time to address most of the issues we discussed.
Main changes:
- The dependency on astropy.units is now optional:
   - Any Python property with a _um, _ms, etc. suffix is automatically converted to an MM property with -us, -ms, etc. suffix. 
   - If astropy.units is installed, bootstrap.py wraps properties with a Quantity[u.um], etc. return type to behave exactly as the suffixed properties. Units are added and converted as appropriate.

- Major rehaul of bootstrap.py:
   - The 'Protocol' and 'isinstance' approach is removed completely. Instead, bootstrap.py now investigates all properties (including property type) and methods, and constructs an object PyDevice
   - The PyDevice object contains a list of properties (with getters and setters), and methods. All getters/setters/methods are curried, meaning that they can be called without reference to the original object. This way, the c++ code does not need to keep track of a reference to the device object. Also, it allows for greater flexibility because we can easily wrap the properties and methods, such as is done for handling astropy units. Also, in the future this mechanism could be used to construct a single PyDevice that combines properties and methods of multiple Python objects, or allow the user to modify the list of properties/methods manually.
   - Easier to run independently. Python code can just call the `PyDevice` constructor to inspect an object. This makes it easier to see if all properties are converted correctly and if the device type is recognized correctly.

- Added some pytest tests for testing bootstrap.py
- Added some tests for loading and using PyDevices in `pymmcore`. The tests are not extensive, but it does work so far.
- Major cleanup of the c++ code.  

There is still a lot to do, such as:
- Updating the documentation
- Fixing the stages/homing (can anyone give a clear documentation of how this is supposed to work?)
- Update to use the new frame grabbing api (any documentation here?)
- More testing, and getting feedback from some alpha users.
- Documenting the numpy dependency. In the future we could move to the buffer protocol so that numpy is not needed anymore. However, this protocol is only present in the stable API since Python3.11. But if we wait 2.5 years, Python 3.10 is end of life...   

I'll update the README.md in the next couple of days and let you know when I am done. After that, can we merge the PR and advertise it as 'work in progress'?

marktsuchida:
@IvoVellekoop That's awesome! I'll plan to look at this this week and get ready to merge it.

IvoVellekoop:
Dear all,

I managed to change the process of loading the Python runtime so that virtual environments are fully supported, and there is no need to have a 'global' Python installation or set the PATH variable. 

The technical implementation was quite hard, mainly because initializing the Python runtime is insanely complicated and poorly documented. The Python dll is now delay-loaded, meaning that PyDevice can select the correct version if multiple runtimes are installed. This also has the benefit that the PyDevice plugin is still visible in Micro-Manager even if Python is not in the system PATH. In addition, the new buffer protocol is now supported for all Python versions from 3.9 onwards, so there is no longer the need to install numpy, meaning that PyDevice runs on vanilla Python. Devices may also, for instance, return Pillow or pytorch data arrays, although I haven't tested that yet.

I've updated the documentation to reflect all recent changes. @marktsuchida, @tlambert03 and @henrypinkard, I think the device is now ready for additional testing and then merging. It should be reasonable future proof now, also with the re-designed bootstrap.py code.

Looking forward to your comments,

Ivo

henrypinkard:
Awesome! I'll try to look at this in detail before the end of the week

DeviceAdapters/PyDevice/PyCamera.cpp
```diff
@@ -0,0 +1,252 @@
...
+* Sets binning factor.
+* Required by the MM::Camera API.
+*/
+int CPyCamera::SetBinning(int binF)
+{
+    return SetLongProperty(g_Keyword_Binning, binF);
+}
+
+int CPyCamera::IsExposureSequenceable(bool& isSequenceable) const
+{
```
henrypinkard, line 224: This should default to False. Exposure sequencing is an advanced feature that most cameras do not support

henrypinkard:
I've looked through again in more detail. Looks like you've a made a lot of good changes! Below is this list of things I found that could use correction, that I don't quite understand, and/or that need more testing/discussion. Most of this I think can be addresses in future PRs, since it becomes hard to keep track of so many things on a single monolithic PR.

I suggest we proceed with merging this PR and releasing the alpha version as soon as @marktsuchida confirms readiness. 
After merging, if there are no objections, I'd say the next step is to create a new repo within the micro-manager organization (clearly marked as experimental and unstable), to house discussions and more development of these features.

# Feedback from Testing

- Consider moving pydevice.md from the examples folder to the docs folder.

- There seems to be a bug with some of the example files -- when they are run from the hardware configuration wizard when adding the pydevice, `__file__` is not found leading to this error:
Screenshot 2024-04-24 085349
It works fine when commented so I think this just needs a check for None

- If you change the ScriptPath in the hardware config without adding/deleting the hub device, it doesn't seem to update what peripheral devices are available (its possible this is a bug on the MM side)

- The current setup involves creating a dictionary of devices in Python, which adds an extra layer of configuration outside of MM. It might be more robust to integrate this information directly into the hardware config. For instance, if a microscope.py file creates this dictionary, is somewhat obscures what devices are available. Perhaps using multiple scripts, one per file/device, could simplify this setup?

- If I understand correctly, one hub device corresponds to one python interpreter. Is it possible to multiple hubs/interpreters? For example, if there were an inefficiently written python device that is constantly holding the GIL, could making multiple python devices be a way around this?

- I noticed there is now a fairly complicated hierarchy for finding which python interpreter to use (https://github.com/micro-manager/mmCoreAndDevices/blob/5a1cd54d70ffd2e33872d75293ec2a5844ab5fa4/DeviceAdapters/PyDevice/stable.cpp#L70) I think it would be good to make this as explicit as possible in the docs, since I could see this being a major source of confusion if it goes wrong.

## Python Side Device APIs


- There is a discrepancy between the required properties and methods for Python implementations of devices compared to MM's requirements. In certain cases where MM uses methods, Python uses properties, which could lead to confusion. On the other hand, some things are easier/more natural to express as properties on the python side. So at some point it would be great to hear more about your thinking on this, and I think it would be beneficial to establish a rule for when Python devices should mirror MM's approach versus implementing a simpler alternative.

### Camera device
- Consider renaming `left` and `right` properties in the camera device to contain `"roi"` for naming consistency with the rest of the MM APIs
- The camera.py file provided in the examples does not work, I think because it does not implement the `trigger` method
- Regarding python side camera API, it is quite different from the Camera MMDevice, and it currently only supports snapping image. This is something I suspect that will require a good bit out thought/discussion regarding if and how this should differ from MMDevice, as well as how it fits with the new camera API (https://github.com/micro-manager/mmCoreAndDevices/issues/243), which I plan to move towards finalizing soon.

tlambert03:
just a quick note: you might consider using [`annotated_types.Unit`](https://github.com/annotated-types/annotated-types?tab=readme-ov-file#unit) as a generic way to declare types. `annotated_types` is a minimal library intended to declare (but not enforce directly) constraints in type annotations... it's used by pydantic and others.  It would not enforce anything for you  (you would still have to depend on, e.g. pint to do any actual conversions), but it would provide a minimal-dependency syntax that may ultimately be used by other backend-agnostic libraries.

See https://github.com/annotated-types/annotated-types/issues/64 for discussion.

IvoVellekoop:
@tlambert03 Thank you for the suggestion. I've added support for annotated_types so that you can use Ge, Le, Interval and Unit. 
The modules astropy.units, numpy and annotated_types are all completely optional now, so PyDevice also works with a vanilla Python installation. Could you perhaps have a look at the new bootstrap.py? I'd love to hear your thoughts. Implementing pint support would be a nice and simple extension.

@henrypinkard Thanks for all the feedback. I fixed most of the issues:
* IsExposureSequenceable should return false. ✓
* Moving pydevice.md from the examples folder. This was actually an obsolete file. I removed it. All documentation is now in the README.md. ✓
* __file__ not found. The __file__ variable is now set to 'pydevice' when running from Micro-Manager. ✓
* Complicated hierarchy for finding which python interpreter to use. This was simplified a lot when implementing the DELAYLOAD functionality. It is now also documented clearly (I hope) in README.md. 
* The camera.py file provided in the examples does not work. Fixed, and I verified that all other examples also load correctly. ✓ 


Remaining issues that we should think about a bit more:
* If you change the ScriptPath in the hardware config without adding/deleting the hub device, it doesn't seem to update what peripheral devices are available (its possible this is a bug on the MM side)
>> If the Hub object is destroyed and recreated, it *should* load the new Python script. If this doesn't happen, indeed we need to investigate what is happening exactly.

* The current setup involves creating a dictionary of devices in Python, which adds an extra layer of configuration outside of MM. It might be more robust to integrate this information directly into the hardware config. For instance, if a microscope.py file creates this dictionary, is somewhat obscures what devices are available. Perhaps using multiple scripts, one per file/device, could simplify this setup?
>> We did experiment with several options, and we found it was becoming complicated quite rapidly when different files are used. Especially if the devices must communicate with each other. The philosophy we have now is that all set-up is done in the Python script. This includes 'hidden' stuff, such as making different devices reference each other, setting parameters that we don't want to be vissible in Micro-Manager, or cannot make visible because they have non-supported data types, loading calibration files, etc., etc. The advantage of this approach is that you can use a Python program for setting up the devices in very advanced scenarios that cannot be handled purely in a configuration script. A good compromise may be to add functionality to view and edit the Python script in Micro-Manager directly. 

* If I understand correctly, one hub device corresponds to one python interpreter. Is it possible to multiple hubs/interpreters? For example, if there were an inefficiently written python device that is constantly holding the GIL, could making multiple python devices be a way around this?
>> Technically this is possible if the interpreters use the same runtime and virtual environment. However, is multiple hubs need different virtual environments (possibly with different Python versions) a problem occurs since only one Python runtime can be loaded once at the moment. Changing this would be very complex and error prone.

* There is a discrepancy between the required properties and methods for Python implementations of devices compared to MM's requirements. In certain cases where MM uses methods, Python uses properties, which could lead to confusion. On the other hand, some things are easier/more natural to express as properties on the python side. So at some point it would be great to hear more about your thinking on this, and I think it would be beneficial to establish a rule for when Python devices should mirror MM's approach versus implementing a simpler alternative.
* Consider renaming left and right properties in the camera device to contain "roi" for naming consistency with the rest of the MM APIs
>> Agreed, we should think about the naming. Is there a list of standard property names for the different device types that I can use?

* Regarding python side camera API, it is quite different from the Camera MMDevice, and it currently only supports snapping image. This is something I suspect that will require a good bit out thought/discussion regarding if and how this should differ from MMDevice, as well as how it fits with the new camera API (https://github.com/micro-manager/mmCoreAndDevices/issues/243), which I plan to move towards finalizing soon.
>> Let's discuss this ones the new camera API is implemented. openwfs supports more advanced triggering, asynchronous acquisition and multi-threaded data processing. PyDevice only uses a subset of the openwfs camera API at the moment, but I think it can be expanded relatively easily when the new MM camera API is ready.

marktsuchida:
@henrypinkard Thanks for the nudge. I've already started to add Python to 3rdpartypublic and check the build. Should be ready today.

The .gitignore in question is not in a subdirectory.

marktsuchida:
@IvoVellekoop I notice that stable.cpp is not included in the vcxproj. Is this intentional?

marktsuchida:
Python (3.9, Windows x64) `libs` and `include` are now at 3rdpartypublic `r217` `Python/cp39-win_amd64` (added the version/platform so that we won't have trouble if changing/adding later). (I realize the Python headers are not currently used, but in case we need them in the future.)

Hope you don't mind I pushed 2 commits, one to update this path and another to fix the .gitignore issue.

As far as I'm concerned, the only thing blocking the merge is the stable.cpp file situation (see above). Thanks for your patience with this process.

IvoVellekoop:
@marktsuchida. Good to hear that PyDevice is about to be merged, thanks for the help!

I was not aware of any changes in the global .gitignore. If we made changes, they were not intentional and should indeed be reverted.

As for the path of python39.lib, that makes sense. I've modified the build instructions in README.md accordingly.

stable.cpp was listed as a Resource file rather than a Source file in vcxproj.filters. I fixed it now. There should definitely not be a copy of it in 'enc_temp_folder'. It may be something that Visual Studio did, but it was not intentional. I've removed the file from the repo. 

@henrypinkard PyDevice does not rely on any specific version of Python anymore. However, for building we need `python3x.lib` so that the compiler knows which functions to look for in the delay-loaded dll. `numpy` is not needed at all anymore. The use of `astropy` and `annotated_types` is supported and completely optional.

marktsuchida:
Thanks, @IvoVellekoop! Looking forward to working together more after the merge. Thanks again for your major contribution.

marktsuchida:
It looks like the [CI job](https://github.com/micro-manager/mmCoreAndDevices/actions/runs/9472564178/job/26098139325) needed to pick up this change in the main micro-manager repo is not running (stuck on "Waiting for a runner to pick up this job..." for over 30 min).
Hopefully it's a temporary hiccup on GitHub; if it doesn't fix itself overnight I'll look into it again. If this doesn't show up in today's nightly build, this is why.

marktsuchida:
@henrypinkard It's there, otherwise the build wouldn't work.... The latest revision is 3rdpartypublic r217 (not 3rd_party_public).

```console
$ ls Python/cp39-win_amd64/libs
_tkinter.lib python3.lib  python39.lib
```

henrypinkard:
Huh, weird that I didn't see it. 

Does that trac website where one can view this repo still exist or is the only way to inspect to pull to local?

marktsuchida:
For now, https://valelab4.ucsf.edu/svn/3rdpartypublic/Python/cp39-win_amd64/ (certificate is currently expired).

IvoVellekoop:
Was this resolved now? I've cloned the repo and built from scratch this morning to confirm that only python39.lib is needed. So that file indeed needs to be included in the 3rdpartypublic\Python\cp39-win_amd64 folder of the repo if we want developers to build this code out of the box.
Just let me know if I can help.

marktsuchida:
As far as I can tell, there never was a problem: the files are already in 3rdpartypublic (r217 and later). The automated build is already using those files to produce the PyDevice DLL shipped with the nightly build.

henrypinkard:
I think this must have been a problem with my subversion client, since the file is there. We were able to build a new version of it, though there were some difficulties along the way to getting the runtime version of python to work. I'll make a new post with details shortly

PR #416, opened 2023-12-14
Experimental Meson build for MMDevice and MMCore

marktsuchida:
The immediate benefit is that this allows us to run the unit tests for MMDevice and MMCore, which we've never had running in CI previously (and were never run on Windows at all). See the included GitHub workflow.

This build is not (so far) appropriate for building MMCore for distribution or actual use. That will come eventually.

This has no effect on the existing Visual Studio and Autoconf/Automake builds, which need to be kept maintained.

This is a first step toward micro-manager/micro-manager#1392. It's a little out of order compared to the [plan](https://github.com/micro-manager/micro-manager/issues/1392#issuecomment-1156956436) outlined there (which had MMCoreJ first); I intend to create a separate overview issue (in _this_ repo) for the C++ parts of the build modernization, with some updates to the plan. The reason for adding Meson scripts to MMDevice and MMCore first is so that we don't need to repeat the list of source (.cpp) files in more places than necessary (i.e., MMCoreJ and pymmcore).

For anyone trying it locally: in order to build MMCore you need to copy MMDevice/ into MMCore/subprojects/ before setting up Meson. This is because Meson cannot use subprojects in arbitrary locations (which is a good thing, once you adapt). In our case, we can work around this by copying (or symlinking) projects as needed until we start restructuring Git repositories. Once MMDevice and MMCore have their own repository, Meson can fetch them automatically (or we can use Git submodules).

~Marking as draft until the CI passes.~

### DIFF
--- a/.github/workflows/ci-mmdevice-mmcore.yml
+++ b/.github/workflows/ci-mmdevice-mmcore.yml
@@ -1,0 +1,67 @@
+name: MMDevice/MMCore experimental Meson build & tests
+
+on:
+  pull_request:
+    paths:
+      - MMDevice/**
+      - MMCore/**
+  push:
+    branches:
+      - main
+    paths:
+      - MMDevice/**
+      - MMCore/**
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - windows
+          - macos
+          - ubuntu
+        include:
+          - os: windows
+            runner: windows-2019
+            cxx: cl
+          - os: macos
+            runner: macos-11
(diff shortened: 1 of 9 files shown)


tlambert03:
exciting!

marktsuchida:
Nice, I'd only tested this on Windows and it worked the first time!

GCC and Clang are generating a bunch of warnings; I'll address these in a separate PR (in the code rather than the build, since we get the same warnings in our existing build, too).

marktsuchida:
I'll also note that there may be missing symbols once we actually use these from MMCoreJ/pymmcore.

henrypinkard:
Awesome!

marktsuchida:
Thanks, @tlambert03 and @henrypinkard.

I'm going to go ahead and merge this since nothing here is difficult to change later.

PR #427, opened 2024-01-04
Updated support for Mad City Labs devices.

mclgreg:
-(MicroDrive) Added support for new products.
-(MicroDrive) Updated device initialization.  New algorithm for selecting which device and axis gets selected when a new device adpater is initialized. 
-(Madlib) Fixed a bug with handle acquisition.


marktsuchida:
Thanks!

PR #428, opened 2024-01-05
Add NotificationTester device adapter

marktsuchida:
Pure-software device adapter intended for automated (and interactive) testing of application handling of change notifications.

Initial implementation with property simulator only. Simulation of Z/XY stages, etc., will follow soon.

Draft of docs (to be put on device page once merged):

Device `NTSyncProperty` has `TestProperty` (float) that is never busy (immediately reflects set value). If `NotificationsEnabled` is set to `Yes`, `OnPropertyChanged()` is called from within the property `AfterSet` handler, on the same thread that called into the device adapter.

Device `NTAsyncProperty` has `TestProperty` (float) which distinguishes the target value from the current value, and the current value is slewed toward the target at a constant rate (`SlewTimePerUnit_s`), updating periodically (`UpdateInterval_s`). The last update is guaranteed to match the target value (even if that happens mid-update-interval). `Busy()` returns true when the current value differs from the target value. Reading the property always returns the last-updated current value (as opposed to the target). If the property is set before the previous slewing completed, the previous target is forgotten and a slew toward the new target is begun. When `NotificationsEnabled` is `Yes`, `OnPropertyChanged()` is called on every update, after an optional, independent delay (`NotificationDelay_s`), always from a thread owned by the device adapter. A device that only notifies once per change can be simulated by setting the update interval to be longer than the total slew time.

It is hard to know how well this captures the behavior of every existing device adapter, although the intent is to capture most possibilities that have some degree of coherence. I'm sure there are devices that have other quirks that cannot be simulated with this scheme (some of those may be bugs, but perhaps not all).

### DIFF
--- a/DeviceAdapters/Makefile.am
+++ b/DeviceAdapters/Makefile.am
@@ -155,6 +155,7 @@ SUBDIRS = \
 	NewportSMC \
 	Nikon \
 	NikonTE2000 \
+	NotificationTester \
 	OVP_ECS2 \
 	Omicron \
 	Oxxius \
--- a/DeviceAdapters/NotificationTester/DelayedNotifier.h
+++ b/DeviceAdapters/NotificationTester/DelayedNotifier.h
@@ -1,0 +1,137 @@
+// Mock device adapter for testing of device change notifications
+//
+// Copyright (C) 2024 Board of Regents of the University of Wisconsin System
+//
+// This file is distributed under the BSD license. License text is included
+// with the source distribution.
+//
+// This file is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+// FOR A PARTICULAR PURPOSE.
+//
+// IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY
+// DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES.
+//
+// Author: Mark A. Tsuchida
+
+#pragma once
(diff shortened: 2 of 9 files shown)


marktsuchida:
- [ ] `delayer_` needs to be flushed in `Shutdown()`
- [ ] Avoid creating unused `DelayedNotifier` thread in sync device (lazy-create on schedule?)
- [ ] Add second property `ExternallySet` that sets the target of `TestProperty` (and itself reads as written)

marktsuchida:
(More doc drafts)

For both devices, setting the `ExternallySet` property will simulate the target value being changed from the hardware side. The `ExternallySet` property itself is always read as last written.

`NTSyncStage` and `NTAsyncStage` work similarly, except that busy status and notifications apply to the stage position instead of a test property. Notifications are via `OnStagePositionChanged()`. These devices have `ExternallySetSteps` instead of `ExternallySet` and `SlewTimePerStep_s` instead of `SlewTimePerUnit_s`. The step size is fixed at 0.1 um.

marktsuchida:
- [ ] Refactor process models to use `std::array<double, N>` (N = 1 or 2) instead of `double`
- [ ] Add simulated XY stage

marktsuchida:
`NTSyncXYStage` and `NTAsyncXYStage` are the analogous devices simulating XY stages. Notifications are via `OnXYStagePositionChanged()`. The `ExternallySetSteps` property takes a string consisting of two integers separated by `;` (semicolon). The step size is fixed at 0.1 um for both X and Y. For `NTAsyncXYStage`, when the position is set, each of the X and Y axes is slewed at the same rate toward the setpoint; if one axis reaches the setpoint first, it stops first and the other axis keeps slewing.

The Z and XY stages do not support homing but do support stopping.

PR #441, opened 2024-02-06
Made Device Adapter compatible with the multi-camera utility

Lars-Kool:
Hi Mark,

After reading the comments on the image.sc forum. I took a look into multi-camera setups, and made the IDS camera device adapter compatible with the Utilities->multi camera adapter.

Best,
Lars

### DIFF
--- a/DeviceAdapters/IDSPeak/Changelog.txt
+++ b/DeviceAdapters/IDSPeak/Changelog.txt
@@ -6,3 +6,6 @@ Changelog IDSPeak:
 2023.10.10
 - Changed .dll to Release version (no longer requires MS Visual Studio to work)
 - Added multi-camera support
+
+2024.02.06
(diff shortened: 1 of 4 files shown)


marktsuchida:
@Lars-Kool Great!

It looks like your commit also deletes the .vcxproj and .vcxproj.filters files. I'll fix that before merging this (unless you get to it first).

Lars-Kool:
Ahh yes I did. I thought I mistakenly added them to the previous pull request, hence I removed them. But they are back now.
Lars

marktsuchida:
Thanks, @Lars-Kool! I'm force-pushing to your branch after rebasing on main and squashing your 2 commits so that the delete/readd doesn't clutter the history. Also added another commit that removes an unused include (doesn't compile with it).

PR #443, opened 2024-02-07
Linux Spinnaker device adapter.

HazenBabcock:
This will build the Spinnaker device adapter on linux system.

I had to make a few changes to SpinnakerCamera.cpp and SpinnakerCamera.h that I don't think will have an effect on the Windows build, but I'm not 100% sure of this.


marktsuchida:
Thank you for this!
The code changes look safe for Windows but I needed a minor change to the project for it to compile (pushed to your branch).
On Linux I only checked that the build works in the absence of Spinnaker SDK. Thank you for adding the conditional configuration.

HazenBabcock:
Looks good. I checked and it works for me on Linux with the Spinnaker SDK. Thanks!

PR #444, opened 2024-02-22
Bug when controlling CoolLED pE2

alejandrcastro:
We tried to address a bug in precisExcite.cpp that only allowed to change the last channel of the CoolLED. 

Left debugging output code inside, hoping other people might test it too (see https://forum.image.sc/t/cant-turn-on-coolled-pe-2-from-micromanager-2-0/89702/7). 

This now seems to work with our system.

### DIFF
--- a/DeviceAdapters/PrecisExcite/PrecisExcite.cpp
+++ b/DeviceAdapters/PrecisExcite/PrecisExcite.cpp
@@ -33,6 +33,22 @@
 #include "ModuleInterface.h"
 #include "DeviceUtils.h"
 #include <sstream>
+#include <iostream>
+#include <fstream>
+
+
+int debug(std::string text) {
+    
+   /*
+   * //LogMessage(text); use LogMessage within the class for better results.
+   * 
+    ofstream myfile;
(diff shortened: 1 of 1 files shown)


marktsuchida:
Thanks for the fix!

If I'm reading this correctly, the actual change is to the `C?` command being sent in `GetState()`: previously it was getting the on/off of all channels and checking if any were on; now it just checks the current channel. This seems correct given how `state_` is being handled around where `currentChannel_` gets changed.

alejandrcastro:
Yes, we assumed that since only one channel is active at any given time, there is no need to check if all channels are on. 
The fact that it was checking all channels lead to only channel D staying on after changing the power, i.e. if channel B was on, and we changed channel B power, the looping through channels would realise channel D was off, and turned everything off. 
Only when D was the active channel, were we able to change the power output and not have it go off. 

It now seems to work to some extent, a user in the imagesc forum thread linked abouve reported it now working in one of his setups but not in another. At this point I should add: precisExcite it still is quite slow unfortunately.

marktsuchida:
Thanks for filling in the details!

PR #446, opened 2024-03-05
Galvo projector: adds ability to run polygons to Utilities::DAGalvo device.

nicost:
There is ample ambiguity who controls the shutter and channel config while photo-targetting, the UI code or the device adapter.

We should discuss the best path forward, but in the mean time this works.

### DIFF
--- a/DeviceAdapters/Utilities/DAGalvo.cpp
+++ b/DeviceAdapters/Utilities/DAGalvo.cpp
@@ -34,6 +34,7 @@
 
 #include <chrono>
 #include <thread>
+#include <vector>
 
 
 extern const char* g_NoDevice;
@@ -46,6 +47,7 @@ DAGalvo::DAGalvo() :
    pulseIntervalUs_(100000),
    initialized_(false)
 {
+   polygons_ = new std::vector<DAPolygon*>();
 }
 
 DAGalvo::~DAGalvo()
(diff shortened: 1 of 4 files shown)


DeviceAdapters/Utilities/DAGalvo.cpp
```diff
@@ -262,8 +283,14 @@ int DAGalvo::SetPolygonRepetitions(int repetitions)
 
 int DAGalvo::RunPolygons()
 {
-   return DEVICE_NOT_YET_IMPLEMENTED;
+   for (int i = 0; i < polygons_->size(); i++) {
+      DAPolygon* polygon = polygons_->at(i);
+      std::pair<double, double> xyPair = polygon->getVertex(0);
```
marktsuchida, line 288: Did you mean `getVertex(i)`?

marktsuchida: Never mind, I guess this is intentional(?)

marktsuchida:
What is the reason for disabling the `SleepMs(1)` in `SnapImage()` on Windows?

nicost:
Yikes, that is a left over from trying to get those sleeps to sleep shorter.  I'll add another PR to bring this back.

DeviceAdapters/Utilities/DAGalvo.cpp
nicost: Yup, so far it only targets points, simplified by only using the first entry of the polygon.  Could be extended in the future to draw the complete polygon.

Sorry for merging so fast!  I could have avoided some problems by waiting.

PR #451, opened 2024-03-13
Move `onPropertyChanged("Core"`  calls to their respective setters

tlambert03:
currently, if you. call `core.setProperty("Core", ...)` you will get a `onPropertyChanged("Core"` callback.  However, if you directly call the setter method called by `CorePropertyCollection::Execute`, the end result is the same, but you don't get a callback.  This moves the callbacks inside the setter methods.  It's unfortunately more code, but it makes callbacks more consistent with internal state.

### DIFF
--- a/MMCore/MMCore.cpp
+++ b/MMCore/MMCore.cpp
@@ -804,6 +804,9 @@ void CMMCore::unloadAllDevices() throw (CMMError)
       logError("MMCore::unloadAllDevices", err.getMsg().c_str());
       throw;
    }
+   if (externalCallback_) {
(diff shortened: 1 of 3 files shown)


MMCore/MMCore.cpp
```diff
@@ -4225,6 +4259,17 @@ double CMMCore::getExposure(const char* label) throw (CMMError)
      return 0.0;
 }
 
+
+void CMMCore::setTimeoutMs(long timeoutMs)
```
tlambert03, line 4263: :eyes: @marktsuchida  did i move this out of MMCore.h correctly?

marktsuchida:
Note to self: should these be asynchronous callbacks?

PR #454, opened 2024-03-28
Add support for Player One camera and filter wheel

PlayerOneAstronomy:
Hello, maintainer, I have added support for Player One camera and filter wheel in DeviceAdapters. Features development have been completed and tested, and it work well in Micro-Manager 2.0.3.
Thanks.


nicost:
Sorry for the delay in working on this. I tried to compile the code, but found the following dependencies missing:
- ConvFuncs.h
- PlayerOneCamera.h
- PlayerOnePW.h
There likely are also .lib files needed for linking missing.  

We usually put such files in our 3rdpartypublic repository. If you can point me to these files, and their license, I'd be happy to do so.

PlayerOneAstronomy:
Hi @nicost , Thank you very much for your reply.
Here is the headers and libs, and sorry, I put them in the '3rdparty' path, not 3rdpartypublic.
[PlayerOne.zip](https://github.com/micro-manager/mmCoreAndDevices/files/14976354/PlayerOne.zip)

nicost:
Sorry for the delay.  The device adapter should be in the next nightly build.  Could you please check that everything work correctly?  Thanks for your patience!

PlayerOneAstronomy:
> Sorry for the delay. The device adapter should be in the next nightly build. Could you please check that everything work correctly? Thanks for your patience!

Thank you nicost! I will do the check and test.

PR #467, opened 2024-06-17
Fix Scientifica device adapter - Motion 2 Y moves moving incorrect distance

MatthewPlayer:
Fix bug where using the Scientifica device adapter with Motion 2 racks for Y axis moves, move 10 times less than they should.


marktsuchida:
Thanks!

PR #469, opened 2024-06-19
Fix character encoding in source files

marktsuchida:
All to UTF-8.

GitHub appears to be quite good at guessing the encoding, so many of the diffs may look like they are not changing anything (which is a good thing).

TODO:

- [x] Update vcxproj source encoding (if necessary)
- [x] Enforce UTF-8 for `.h`, `.cpp` in CI


marktsuchida:
I've taken care to make sure that these changes don't break anything on Windows (details in commit messages); if anything this may have fixed some non-ASCII strings on Windows computers set to a non-latin1 language. Non-Windows machines are probably not affected. But it's always possible that I forgot to consider some case, and it's hard to test these things. Fingers crossed.

PR #472, opened 2024-06-26
FRE / excelitas pco gmbh / 26.06.2024 / added 48bit RGB

franz-xaver:
Changed MicroManager.cpp in pco_generic device adapter.

FRE

### DIFF
--- a/DeviceAdapters/PCO_Generic/MicroManager.cpp
+++ b/DeviceAdapters/PCO_Generic/MicroManager.cpp
@@ -45,6 +45,7 @@ const char* g_CameraDeviceName = "pco_camera";
 const char* g_PixelType_8bit = "8bit";
 const char* g_PixelType_16bit = "16bit";
 const char* g_PixelType_RGB32bit = "RGB 32bit";
+const char* g_PixelType_RGB48bit = "RGB 48bit";
 
 const char* g_TimeStamp_No = "No stamp";
 const char* g_TimeStamp_B = "Binary";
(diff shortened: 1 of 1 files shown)


franz-xaver:
FRE

marktsuchida:
@franz-xaver Thanks for the update.

I don't think we have standardized or documented how 16-bit-per-component RGB is supposed to work, and there are existing camera adapters that use an 8-byte (rather than 6-byte) format. Please give me a bit of time to review the situation to make sure that merging this will not produce compatibility problems in the future.

(I do hope that we can safely allow 48-bit RGB, just want to be sure.)

By the way, did you test this using MMStudio or some other way that goes through MMCore? If so, let me know how so that I can make sure to take that usage into consideration.

Also, does the buffer that you pass to `InsertImage()` contain the pixel components in R-G-B order (as opposed to B-G-R order)?

franz-xaver:
@marktsuchida ...sorry for the last commit.

The title of the commit is not well verbalized. It is more a preparation for future things to come.

A customer asked me to add this image type into MM, but I did without success.

The code is just added but not active, as I could not add a 16bit RGB image:

```  if (m_pCamera->GetCCDCol(0))
  {
    pixTypes.push_back(g_PixelType_RGB32bit);
    //pixTypes.push_back(g_PixelType_RGB48bit); // Not possible up to now 07/22
    // Also GetImage does not convert to 48bit rgb, just 32bit rgb
  }```

Actually this is untested and RGB is not the order. It is just a naming. RGB vs. BGR should be added.

franz-xaver:
Yes, correct.

PR #474, opened 2024-07-01
Expose trigger polarity to Micro-Manager in Point Grey device adapter

kmdouglass:
This change exposes the trigger polarity to Micro-Manager (if it exists). The polarity determines whether an external trigger starts an exposure when it is "high" or "low".

Co-authored with @GiorgioTortarolo


marktsuchida:
Thanks!

PR #477, opened 2024-07-05
Revert pull request #476

marktsuchida:
The check for an initialized hub was already in HubInstance.

@tlambert03 Sorry I missed this while reviewing #476. See
https://github.com/micro-manager/mmCoreAndDevices/blob/36c45e604a6a58dd08427c542881e029d419c5a1/MMCore/Devices/HubInstance.cpp#L36
which is in the code path of `getInstalledDevices()`.

Is it possible that the above check is not functioning (with the `StrictInitializationChecks` feature enabled)? If so, let me know how to reproduce. It should also log a warning with the feature disabled.


tlambert03:
Oh good catch!  Sorry I missed it too!  

I actually think it's more likely I hadn't set the feature flag in my usage than it is that it's not functioning. 

Revert is fine, sorry for the noise

marktsuchida:
Not a problem!

PR #478, opened 2024-07-07
OpenFlexure Device Adapter

samdrea:


marktsuchida:
Doesn't build. Missing license. @henrypinkard I'd appreciate it if you gave me (us Core Devs) at least a day or two to review new device adapters.

The build failure is due to a PostBuildCommand in the .vcxproj. I will fix this shortly in a separate PR.

Please add a license -- BSD or LGPL.

On a separate note, wonderful to see OpenFlexure supported in MM!

marktsuchida:
Build fixed in #479. Should go in the next nightly build.

henrypinkard:
Ah my mistake -- will do in the future!

henrypinkard:
Btw -- do you have a strategy for copying in DLLs locally after builds that's automated but doesnt break main build due to the post build events

marktsuchida:
No problem, @henrypinkard.

People leaving in local development settings is actually one of the more common things I find myself fixing for new device adapters (and sometimes existing ones).

> do you have a strategy for copying in DLLs locally after builds that's automated but doesnt break main build due to the post build events?

Well, it shouldn't be in committed code, so I would try something that doesn't require modifying the project files. Personally I just have a `cp` command in Git Bash that I recall and run, but if you want it to be one step, you could do it by invoking the build through MSBuild (this must be done in the Developer PowerShell for Visual Studio 2019):

```pwsh
cd path\to\mmCoreAndDevices\DeviceAdapters\DemoCamera
msbuild DemoCamera.vcxproj /p:Configuration=Debug /p:Platform=x64 && cp build\Debug\x64\mmgr_dal_DemoCamera.dll path\to\Micro-Manager-2.0\
```

(Note that the build products go into `.\build` rather than `mmCoreAndDevices\build` in this case. IIRC it's relative to the solution, but here we're not using the solution.)

The defaults are `Debug` and `Win32`, so you need to set at least the `Platform`.

henrypinkard:
good to know, thanks!

PR #429, opened 2024-01-10
support new camera

fandayu:
support aries 16 lt/fl 9bw/fl 26bw

PR #482, opened 2024-08-20
Update TUCam plugin code

fandayu:
Upgraded the project

### DIFF
--- a/.github/workflows/ci-mmdevice-mmcore.yml
+++ b/.github/workflows/ci-mmdevice-mmcore.yml
@@ -30,7 +30,7 @@ jobs:
             runner: windows-2019
             cxx: cl
           - os: macos
-            runner: macos-11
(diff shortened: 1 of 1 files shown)


marktsuchida:
@fandayu Unfortunately it looks like this pull request's branch is based on an old pull request of yours, #429, instead of the current `main`, which has had a few changes to TUCam since #429. This is why it says that it has conflicts.

Can you make your changes to the current `main` version ([here](https://github.com/micro-manager/mmCoreAndDevices/tree/main/DeviceAdapters/TUCam))? Please make sure to also build and test using the up-to-date `MMTUCam.vcxproj`, because some settings have changed.

(Note that your fork's `main` is not the same as the official (upstream) `main`, because you have committed past changes to it.)

It might be easier for you to create a new pull request and close this one.

PR #485, opened 2024-08-22
ASITiger: fix missing semicolon

marktsuchida:
Cc: @bls337, in case you want to re-check this code.


bls337:
@marktsuchida Thanks!

PR #488, opened 2024-09-03
Adding autofocus, general improvements

martinzak-zaber:
We are adding autofocus support for our Nucleus™️ microscope.

There are also a few general improvements:

- Support for network sharing
- Resetting connection on port change
- Adding `override` to all the API methods

We are also updating to a newer version of our library - 3.4.4. Can we please have SVN updated with our binaries? Please note that the tag.gz contains soft-links. Archive: [https://drive.google.com/file/d/1IisP6pyjYEEDqVsKLyruIEPc5TLhDuW7/view?usp=sharing](https://drive.google.com/file/d/1IisP6pyjYEEDqVsKLyruIEPc5TLhDuW7/view?usp=sharing)

Note: We have tried upgrading our module to C++17 but could not get it working without removing [dynamic exceptions](https://en.cppreference.com/w/cpp/language/except_spec) from some of the base classes.

Thank you. Please let us know if you have any concerns or questions.

### DIFF
--- a/DeviceAdapters/Zaber/ConnectionManager.cpp
+++ b/DeviceAdapters/Zaber/ConnectionManager.cpp
@@ -1,7 +1,10 @@
 #include "ConnectionManager.h"
 
+#include <regex>
+#include <zaber/motion/exceptions/invalid_argument_exception.h>
+
 std::shared_ptr<zml::Connection> ConnectionManager::getConnection(std::string port)
-{	
+{
 	std::lock_guard<std::mutex> lockGuard(lock_);
 	if (connections_.count(port) > 0) {
 		if (auto connectionPtr = connections_.at(port).lock()) {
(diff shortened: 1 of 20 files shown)


marktsuchida:
Thanks, @martinzak-zaber!

- I added the new ZML as `3rdpartypublic/Zaber/zaber-motion-3.4.4`, keeping the previous `zaber-motion` (3.4.3). (We do this so that the 3rdpartypublic revision doesn't have to be in exact lock step with this repo.)

- I pushed a commit to your branch (this PR) that updates the build files accordingly.

- Regarding C++17 (for future reference), we have a precedent on Windows (PyDevice) that currently builds with C++17. On Linux (Automake), I think the following (in `Zaber/Makefile.am`) should work and suppress the warnings about the dynamic exception specifications (but did you find otherwise?).

```make
AM_CXXFLAGS = $(MMDEVAPI_CXXFLAGS) $(ZML_CPPFLAGS) -std=c++17 -Wno-dynamic-exception-spec
```

martinzak-zaber:
@marktsuchida We have been able to build on Windows as well. On Linux we got hard error that dynamic exceptions are not supported in c++17 along with "-Wno-dynamic-exception-spec" being an unknown option. Maybe we had a wrong GCC version but I could not find any documentation on "no-dynamic-exception-spec" anyway so I gave up.

marktsuchida:
@martinzak-zaber You're right, sorry about that. I must have misremembered the situation back when I tried that flag; it turns out that only [Clang](https://clang.llvm.org/docs/DiagnosticsReference.html#wdynamic-exception-spec) has `-W[no-]dynamic-exception-spec`.

#489 should fix this so that you just need `-std=c++17` in `AM_CXXFLAGS` next time you update.

martinzak-zaber:
> @martinzak-zaber You're right, sorry about that. I must have misremembered the situation back when I tried that flag; it turns out that only [Clang](https://clang.llvm.org/docs/DiagnosticsReference.html#wdynamic-exception-spec) has `-W[no-]dynamic-exception-spec`.
> 
> #489 should fix this so that you just need `-std=c++17` in `AM_CXXFLAGS` next time you update.

Much appreciated. Thank you!

nicost:
@martinzak-zaber This code currently does not build on LInux.  Shall I just completely disable the Zaber build on Linux (which would be a shame)?  We do not have the 3.4.4 motion library linux .so files in our 3rdparty repo.  I can not check the .tar.gz you posted earlier since it disappeared. 

Let me know how to proceed.  For the time being, I will disable the Zaber build on linux (as it does not work, and blocks the build of everything else).

marktsuchida:
@nico The Zaber headers are in 3rdpartypublic and the configure.ac should disable the device adapter when they are not present (and the correct version). If this is not the case, can you create a new issue with details? I'll look into a fix.

martinzak-zaber:
@nicost @marktsuchida Hello.

I am sorry for missing the mention notification. I've just come back to this thread because I've also failed to build our plugin on Ubuntu. Could you please update the 3partypublic and include the linux-amd64 binaries?

The easiest way would be to unpack them (`dpkg-deb -R`) from our deb distribution: [v3.4.4](https://www.zaber.com/support/software-downloads.php?product=zml_cpp_linux64_deb&version=3.4.4). The binaries are in /usr/lib folder in the deb package. Please note that both the binary and the symlinks are needed. I verified that the build passes.

Does this work? I can also create another tar.gz.

Thank you for maintaining MicroManager.

martinzak-zaber:
Thank you, @marktsuchida. Everything works on our side, too.

PR #497, opened 2024-09-23
LightSheetManager: use C++17 in debug and release mode

bls337:


marktsuchida:
@bls337 Would it be okay if we left it set to C++14 until the code actually requires C++17?

In part because we'll also need to enable C++17 for the Mac/Linux build, and it will be good to test all when merging the code that uses C++17 features (there are no other device adapters using C++17 on Mac/Linux yet, even though I did do a preliminary test that it should work).

Same for #495 and #496 as well.

(You are welcome to start using C++17 features whenever you like.)

bls337:
@marktsuchida That works for me, I was curious about how much of a pain it would be to enable right now. If it's easier to just leave it at C++14 for now we should do that. I will enable it if I take advantage of the new C++ features.

marktsuchida:
@bls337 Thanks. I don't expect it to be a pain, but it will be good to keep the build scripts in sync.

PR #499, opened 2024-10-09
Fix two typos

HazenBabcock:
This replaces `PrepareSequenceAcqusition` with `PrepareSequenceAcquisition` and `g_Keyword_Meatdata_Exposure` with `g_Keyword_Metadata_Exposure`. It touches a lot of files because the first typo is present in most of the camera device adapters.

### DIFF
--- a/DeviceAdapters/AmScope/AmScope.cpp
+++ b/DeviceAdapters/AmScope/AmScope.cpp
@@ -694,7 +694,7 @@ int AmScope::SetBinning(int binF)
    return SetProperty("Binning-Software", CDeviceUtils::ConvertToString(binF));
 }
 
-//int AmScope::PrepareSequenceAcqusition()
+//int AmScope::PrepareSequenceAcquisition()
 //{
 //   if (IsCapturing())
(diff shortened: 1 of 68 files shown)


tlambert03:
Hey @HazenBabcock 👋, this is a duplicate of #453  ... see discussion there for hesitation on merge

HazenBabcock:
Yes it is, apologies for that.. I should have checked the open pull requests first..

marktsuchida:
No problem, we appreciate this!

I think the `Meatdata` will be better to fix by keeping the old name (deprecated) for now, so as not to break things that depend on MMCore (including pymmcore), which exposes it in its API.

HazenBabcock:
I could make another pull request where `Meatdata_Exposure` and 'Metadata_Exposure` are both set? Then once `Meatdata_Exposure` is fixed in the dependencies it could be removed?

marktsuchida:
@HazenBabcock Yes, that would be perfect.

HazenBabcock:
Closed, see #503.

PR #514, opened 2024-10-24
NIDAQ: Added analog input functionality

SAJHCamps:
Expanded the driver to be able to read analog inputs and capture traces. Has been tested with NI MAX emulation software but not with actual hardware.

Things to check before merging:
- [x] Run tests with actual hardware.
- [x] Currently the trace values are read out one sample at the time but I am unsure whether this is a mistake on my side or a side effect of the emulator. 
- [x] Check whether the NIDAQmx commands used are available in the 9.2 version of the library.

### DIFF
--- a/DeviceAdapters/NIDAQ/NIAnalogInputPort.cpp
+++ b/DeviceAdapters/NIDAQ/NIAnalogInputPort.cpp
@@ -111,11 +111,15 @@ int NIAnalogInputPort::SetRunning(bool open)
 {
     if (open && !running_)
     {
-        GetHub()->StartAIMeasuringForPort(this);
+        int err = GetHub()->StartAIMeasuringForPort(this);
+        if (err != DEVICE_OK)
+            return err;
     }
     else if (!open && running_)
     {
-        GetHub()->StopAIMeasuringForPort(this);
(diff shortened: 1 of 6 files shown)


nicost:
Your branch compiles fine with the 9.2 version of NIDAQms that we are building against.  There are a couple of compiler warnings, but those will be easy to fix before merging.  Let me know if it is ready for merging.

SAJHCamps:
I just went through the compiler warnings and the ones that are left fall into two categories:
- type conversions, because functions relating to device properties want doubles and NIDAQxm wants float64. Similarly NIDAQxm wants UInt64 but size() for vectors returns t_size.
- unreferenced formal parameters, CSignalIOBase requires several functions which an input port cannot satisfy like StartDASequence(), so DEVICE_UNSUPPORTED_COMMAND is returned without using the parameter.

The NI DAQ card arrived last week, so I expect the other two tasks to be checked off somewhere this week

SAJHCamps:
I ran some test traces and measurements with our PCIe 6323 card and with the fixes the traces and continuous sampling are working as intended. Let me know if there are any more changes needed before merging.

PR #525, opened 2024-11-10
Aravis/pkgconfig

HazenBabcock:
This uses pkgconfig to find the Aravis library, so this is no longer hard coded. However it still only looks for the aravis-0.10 library.


HazenBabcock:
It might be possible to find other versions of Aravis? I was basically cargo-culting from some of the other libraries that are included in MM. If you compile Aravis yourself it creates the `aravis-0.10.pc` file instead of a generically named file like `aravis.pc` and I don't know how to get pkgconfig to search more broadly.

marktsuchida:
Thanks for digging into those Autoconf macros I wrote many years ago!

PR #530, opened 2024-11-30
make names const in mmeventcallback

tlambert03:
This pull request includes changes to the `MMEventCallback` class in the `MMCore/MMEventCallback.h` file to improve the handling of string parameters in callback methods. The most important change is the modification of the type of the `name` parameter from `char*` to `const char*` in several methods.


marktsuchida:
Thanks, should have been this way all along.

PR #532, opened 2024-12-01
implement std::exception for MetadataError

tlambert03:
this adds a `MetadataError::what()`, and adds an overloaded `MetadataKeyError(const char* key)` constructor that will include the missing key in the exception error if provided


marktsuchida:
- [x] This change does not introduce incompatible changes to the MMCoreJ API.
- [x] This change does not affect the device interface ABI.
- [ ] Increment MMCore minor version when merging.

We could also make `MetadataError` an actual subclass of `std::exception` if that helps with anything. Not doing that for now.

marktsuchida:
Had to undo/conditionalize the modern C++ keywords for SWIG 3, which we need for MMCoreJ for now.

Rechecked that there is no spurious change to MMCoreJ API.

PR #535, opened 2024-12-11
New MM::StorageDevice and addition to the MMCore API

go2scope:
Introducing the new MM::StorageDevice. There are two main changes to the micro-manger back end: a new device type (Storage) and the extension of the MMCoreAPI. Both changes are lateral because they extend capabilities without affecting existing functionality. There are no new dependencies for the code base. Any existing client code, including the Micro-manager front end, can safely ignore this extension.

The new feature enables MMCore to save datasets on the back end using a storage device adapter. Developers can add support for various formats by creating device adapters like they would add support for a new camera. There are two significant advantages in utilizing this architecture:

The most important one is that the interface between the microscopy application and the file storage is defined by the MMCore API and allows for designing generic, storage/format-independent acquisition engines and acquisition scripts. To change the file format, we only change the device configuration, but the acquisition code remains the same. Or mainly the same. Adding support for new file formats or using better implementations does not involve changing the existing code.

- By embedding the storage implementation in the MMCore, we can achieve a higher efficiency (data rate), as images do not have to cross the MMCore API boundary. 

I plan on adding more documentation and some examples over the coming days and weeks. This PR is work in progress and only a draft to get some early feedback. Implementation is currently functional, but incomplete. Also, the micro-manager build for the new version may not be completely compatible with the main branch. 

To review the new API, see MMCore.h and MMDevice.h. There are currently two device adapters: BigTiff (with chunking) and Zarr (based on https://github.com/acquire-project/acquire-driver-zarr)


marktsuchida:
Super excited about this coming to MMCore and sorry for the lack of reaction until now!

Would it be okay (before getting into actual code review) if I rebased all your commits to remove the changes that are not direclty part of StorageDevice (mostly the switch of Visual Studio version) and force-push? I'm pretty sure I can do this without losing the branching and merging in your commit history.

This is both (almost certainly) necessary for eventual merging, and should also make the diff a lot easier to view.

(Am right that the code doesn't actually need VS2022?)

(We do want to switch to Visual Studio 2022 very soon, but that really needs to be a separate pull request that is coordinated with build infrastructure updates. That can happen before or after this PR gets merged, depending on what's easiest/needed.)

go2scope:
Hi Mark

The timing is excellent. I have been swamped lately and made little
progress in 2025, but I am on it again. Yes, please, if you know how to fix
the mess I made with my fork and PR, I'd be grateful. I don't care about
branches other than mmstorage-draft.

No, it does not need VS2022. That was a mistake on my part.

I am preparing another PR for the micro-manager repo (the java GUI) to
integrate the new API with the Micro-manager data storage classes. Nico was
helping me, but it turned out to be a more significant challenge for me
than it was to work on MMCore.

Best
Nenad

tlambert03:
was waiting to not be the first person :) ... but also wanted to express my enthusiasm and appreciation for the work being done here.  this would be very exciting to have in some form 🚀

nicost:
Just pushed a commit that reverts the Visual Studio platform toolset to v142 (VS 2019).  Nenad, can you do a git pull before adding other commits? Also, make sure in Visual Studio to select "do not upgrade" if VS offers to upgrade the toolset.  You may need to run the VS installer to install the v142 (VS2019) platform toolsets on your syste.

Regretfully, there are still many project files flagged as changed, at first sight that seems to be due to differences in line ending, so nothing to be really worried about, but it still makes this PR unwieldy.

go2scope:
Hi Nico

Let's talk tomorrow and get that working together. If you copied the zarr
library it should not be grayed out.
Talley recently told me that there is an update to acquire-zarr.dll that
would require changing some code. He did the change on his end, but I
didn't just yet. Maybe the versions don't match.

Acquire zarr is an open source project and can be moved to 3rdpartypublic.

Best
Nenad

marktsuchida:
Before rebasing this branch to remove the VS2022 conversion and other issues with micromanager.sln, I have pushed the current state of the branch to `mmstorage-draft-backup-20250220` in this repository.

marktsuchida:
Now I have force-pushed a rebased branch to the PR branch (mmstorage-draft in the go2scope fork).

Briefly, the rebase

- Removes the conversion of projects to VS2022 (and back to VS2019); the 2 new projects that were newly added are switched to VS2019 in an added commit
- Removes the 2 merges from `main`, which only served to increase conflicts
- Rebases the whole branch onto the current tip of `main` (which could be done without any conflicts in the code!)
- Fixes a build error in MMCore.vcxproj
- Removes `micromanagerSD.slnf` (I don't think this convenience file needs to be in the official repo, and it was causing conflicts)

<details><summary>Full details of the rebase</summary>

After studying (with gitk) the branches and merges within the PR, I found that

- The changes to unrelated vcxproj files are done at the beginning and end of
  the branch, where the histories are linear, and can be simply left out during
  rebasing. (One additional commit, 'reverted to 142 runtime', in the middle,
  can also be left out.)

- Changes to micromanager.sln are more complex, but all of the merge conflicts
  (and their incorrect resolutions) are due to merging `main` into the PR
  branch, which serves no useful purpose. So we can eliminate most problems by
  eliminating the 2 merge commits that merge from `main`. The end result will
  still have conflicts with `main`, but they should be simpler to resolve.

I created a new branch `mmstorage-rebase` equal to the PR branch
`mmstorage-draft` and ran:

```sh
git rebase -i --rebase-merges 4441b057 mmstorage-rebase
```

where 4441b057 is the commit on `main` from which the PR branch began.

The vast majority of the overwhelning amount of commits and branches are
actually branches that have already been merged into `main` (and pulled in
because of the merges from `main` into the PR branch).

Then I deleted the unnecessary lines:

- pick 1b94904b upgrade to VS 2022, solution filter added for Storage Device development
- merge -C 52a0c289 main-13 # Merge branch 'main' of github.com:go2scope/mmCoreAndDevicesSD
- pick f448c836 reverted to 142 runtime
- merge -C 0345b008 main-16 # Merge branch 'main' into mmstorage-draft
- pick 70095d61 Core: revert 0VS platform toolset to v142.
- pick addd0afd Core: add Go2Scope to VS solution.

I then also deleted all the branches upstream of the deleted merge commits.
This was necessary because otherwise all those branches would also get rebased
onto 4441b057, causing conflicts in parts that we are not keeping.

The goal is to rewirte history such that the changes of VS versions and the
merges from `main` never happened.

I anticipated that there would still be conflicts during the rebase at 'zarr
dimension order corrected' (formerly 9f9e03e), because that commit also
included an attempted fix to micromanager.sln (and I could not remove the whole
commit because it also touches source code). This could be resolved by simply
removing the changes to micromanager.sln in this commit.

`git checkout 2efcd502 -- micromanager.sln`

There were also 2 conflicts involving changes to micromanagerSD.slnf, because
that file was created in the removed commit ('upgrade to VS2022, ...'). I take
it that this file is for your personal convenience only, so I resolved the
conflicts by removing the file so it no longer appears anywhere.

After the rebase, I made sure that no commits were missing (except for the
intentionally removed ones listed above), and also examined the diff between
`mmstorage-draft` and `mmstorage-rebase` to ensure that the only differences
are those unrelated to this PR (because I removed the merges from `main`).

After the first round of rebasing, which simplified the whole branch, I did a
second rebase only to edit the commit 'go2scope device introduced' to remove
the spurious (unrelated) changes to micromanager.sln.

In addition, I added a new commit that sets the `PlatformToolset` for the 2 new
projects introduced in this branch to `v142` (VS2019 build tools), so that they
are consistent with all the other projects.

Then I tested the build and found that MMCore does not build, due to
(unintended?) changes to MMCore.vcxproj made in the commit 'static runtime
selected'. So I did a third rebase to remove those changes.

Finally, it (somewhat miraclulously) turned out that there were no code
conflicts with the current `main` (except in micromanager.sln due to projects
added in `main`), so I rebased the whole PR branch onto `main`:

```sh
git rebase --rebase-merges main
```

I reset `mmstorage-draft` to `mmstorage-rebase` and force-pushed it.

</details>

marktsuchida:
We might need to do another round of conflict resolution before merging, depending on the order of events (there's another pending PR that adds device types), but I think this is now in a state where we can review and test -- and I look forward to diving into the details over the next days!

BTW if you want to bring the rebased branch to your local clone, assuming that the remote `origin` points to the go2scope fork (check with `git remote -v`), you can use:

```sh
git fetch origin
git checkout mmstorage-draft
git reset --hard origin/mmstorage-draft
```

marktsuchida:
And don't worry about the failed checks -- I'll fix those before merging, but that's simplest to do just before merging.

MMDevice/MMDevice.h
```diff
@@ -1428,4 +1445,307 @@ namespace MM {
...
+       * \param   [out] handle Output parameter for the dataset handle
+       * \return  Status code indicating success or failure
+       *
+       * \note    - Dimensions are ordered from slowest changing to fastest changing.
+       *          - Typically we are storing images and the last two dimensions are always image height (Y) and width (X)
+       *          - Some implementation may allow non-image data to be stored
+       *          - Declared size for the slowest (first) dimension can be exceeded during acquisition
+       */
+      virtual int Create(const char* path, const char* name, int numberOfDimensions, const int shape[],
+                         MM::StorageDataType pixType, const char* meta, char* handle) = 0;
```
marktsuchida, line 1492: Is there an important reason to use a string handle? Wondering if it might be simpler to use a `uint64_t` handle if the value itself has no meaning.

marktsuchida: Also, for the metadata parameter, I would suggest using a pointer and size (`std::size_t`) rather than a null-terminated string. This would not only eliminate a scan on the receiving side to determine the length, but would also allow metadata that contains internal NUL bytes (not that we'd recommend that).

This applies to all passing of metadata in and out, of course.

MMDevice/MMDevice.h
```diff
@@ -1428,4 +1445,307 @@ namespace MM {
...
+      virtual bool CanLoad(const char* path) = 0;
+
+      /**
+       * \brief Gets the progress of the current operation
+       *
+       * \param handle Dataset handle
+       * \return Progress value (0-100), or -1 if idle or not implemented
+       * \note Assumes single operation execution at a time
+       */
+      virtual int GetProgress(const char* handle) = 0;
```
marktsuchida, line 1584: Is this intended to be called on a second thread while the first thread is making the call to perform an operation?

MMDevice/MMDevice.h
```diff
@@ -1428,4 +1445,307 @@ namespace MM {
...
+
+      /**
+       * \brief   Deletes a dataset
+       *
+       * \details The dataset is permanently removed from storage.
+       * \param   handle Handle of the dataset to delete
+       * \return  Status code indicating success or failure
+       */
+      virtual int Delete(char* handle) = 0;
```
marktsuchida, line 1593: Does this also effectively close the handle? Why does this take a handle rather than a path?

MMDevice/MMDevice.h
```diff
@@ -1428,4 +1445,307 @@ namespace MM {
...
+       * \param   handle Dataset handle
+       * \param   sizeInBytes Size of the image data in bytes
+       * \param   pixels Pointer to the image pixel data
+       * \param   coordinates Array of coordinates for image insertion
+       * \param   numCoordinates Number of coordinate values
+       * \param   imageMeta Image metadata string
+       * \return  Status code indicating success or failure
+       */
+      virtual int AddImage(const char* handle, int sizeInBytes, unsigned char* pixels,
+                           int coordinates[], int numCoordinates, const char* imageMeta) = 0;
```
marktsuchida, line 1618: Sizes should be `std::size_t` or `std::uint64_t` (for memory sizes and on-disk sizes, resp.).
I think coordinates should also be 64-bit, just to be safe. `numCoordinates` can be `std::size_t`.

MMDevice/MMDevice.h
```diff
@@ -1428,4 +1445,307 @@ namespace MM {
...
+
+      /**
+       * \brief   Retrieves image pixel data
+       * \param   handle Dataset handle
+       * \param   coordinates Array of coordinates identifying the image
+       * \param   numCoordinates Number of coordinate values
+       * \return  Pointer to the image pixel data
+       */
+      virtual const unsigned char* GetImage(const char* handle, int coordinates[],
+         int numCoordinates) = 0;
```
marktsuchida, line 1662: Can this have an `int` return value so that it can indicate an error?

Also, how long is the returned pointer valid? If it is not until the handle is closed (which would be problematic), we probably need a "release" function for images (analogous to the one for metadata).

marktsuchida (commented):
I've so far looked at the MMDevice parts. I like the overall design!
Some suggestions and questions.

MMCore/Devices/StorageInstance.cpp
```diff
@@ -0,0 +1,254 @@
...
+int StorageInstance::AddImage(const char* handle, int sizeInBytes, unsigned char* pixels, std::vector<int>& coordinates, const char* imageMeta)
+{
+   RequireInitialized(__func__);
+   return GetImpl()->AddImage(handle, sizeInBytes, pixels, &coordinates[0], (int)coordinates.size(), imageMeta);
+}
+
+int StorageInstance::AppendImage(const char* handle, int sizeInBytes, unsigned char* pixels, const char* imageMeta)
+{
+   RequireInitialized(__func__);
+   return GetImpl()->AppendImage(handle, sizeInBytes, pixels, imageMeta);
```
marktsuchida, line 145: MMCore should throw an error if an attempt is made to add an image to a read-only dataset (again, so that each storage device doesn't have to repeat the logic and we get a consistent error message).

Same with `AddImage` (if keeping).

MMCore/Devices/StorageInstance.cpp
```diff
@@ -0,0 +1,254 @@
...
+         meta = cMeta;
+      GetImpl()->ReleaseStringBuffer(cMeta);
+   }
+   return ret;
+}
+
+int StorageInstance::SetCustomMeta(const char* handle, const std::string& key, const std::string& meta)
+{
+   RequireInitialized(__func__);
+   return GetImpl()->SetCustomMetadata(handle, key.c_str(), meta.c_str());
```
marktsuchida, line 193: Read-only datasets do not allow setting custom metadata, correct?
If so, MMCore should check.

MMCore/MMCore.cpp
```diff
@@ -8032,6 +8042,777 @@ std::vector<std::string> CMMCore::getLoadedPeripheralDevices(const char* hubLabe
...
+
+/**
+ * Pops the next image from the circular buffer and saves it to the dataset.
+ * This is the counterpart to popNextImage(). Instead of fetching image pixels it
+ * it sends image directly to the storage.
+ * 
+ * \param handle - currently open dataset handle
+ * \param imageMeta - image metadata
+ */
+void CMMCore::appendNextToDataset(const char* handle, const std::vector<long>& coordinates, const char* imageMeta) throw (CMMError)
```
marktsuchida, line 8706: Possibly name `popNext[Image]AndAppendToDataset()`, to be easily compared with the plain `popNextImage()` and `appendImageToDataset()`?

---

- This won't work for `Multi Camera`. That can be addressed later (possibly with Henry's per-camera buffering).
- Applications won't be able to store metadata values generated by cameras using this mechanism. This also can be addressed later if found necessary.

(Just noting the limitations so we are aware of them.)

MMCore/MMCore.cpp
```diff
@@ -8032,6 +8042,777 @@ std::vector<std::string> CMMCore::getLoadedPeripheralDevices(const char* hubLabe
...
+   throw CMMError(getCoreErrorText(MMERR_StorageNotAvailable).c_str(), MMERR_StorageNotAvailable);
+}
+
+/**
+ * Returns image pixels at specified coordinates
+ * \param handle - dataset handle
+ * \param coordinates - array of coordinates, one for each dimension
+ * \return - image pixels
+ */
+STORAGEIMGOUT CMMCore::getImageFromDataset(const char* handle, const std::vector<long>& coordinates) throw (CMMError)
```
marktsuchida, line 8609: Just pointing out that this API forces MMCoreJ and pymmcore to immediately make a copy of the image buffer, because the buffer can go away at any moment. (Strictly speaking, even that is not fully safe because the copy happens outside of the device module lock.)

It would be nice to add (in the future) a way to share ownership of the image buffer, so that it can be kept alive as (say) a NumPy array without making a copy. But this need not be part of the first iteration, because the focus is presumably on getting camera-to-storage streaming working. We might want to think about buffer lifetime in the device API, though (I already commented there).

marktsuchida (commented):
Some more comments, primarily on the MMCore API.

marktsuchida:
Thanks, Nenad.

Before further work, I'm going to force-push to this PR branch to remove the 2 commits related to the draft implementation of attached storage: `attachStoragetoCircularBuffer, partial implementation` and `draft implementation of the attached storage`.

The current tip of this PR branch (with those 2 commits) has been kept in the branch [mmstorage-draft-backup-202507](https://github.com/micro-manager/mmCoreAndDevices/tree/mmstorage-draft-backup-202507), so we can resurrect those commits (or at least parts of them) after we finish the basic API and get it merged.

marktsuchida:
@go2scope Can you clarify where the code in 3rdparty/CZI/acquire-zarr (added by you on 2024-11-13) comes from? It doesn't match either acquire-project/acquire-zarr or acquire-project/acquire-zarr-driver, as far as I can tell. For example, I'm having trouble finding the function `ZarrStreamSettings_create()` in (the history of) either of those repos, although equivalent facilities exist.

PR #536, opened 2024-12-12
Add Revealer Scientific Camera Adapter

Ryanyntc2013:
Hello, this is our first time to submit the Micro-Manager plug-in, if you have any questions, please feel free to contact me, thank you very much!

### DIFF
--- a/DeviceAdapters/Revealer/LICENSE
+++ b/DeviceAdapters/Revealer/LICENSE
@@ -1,0 +1,184 @@
+木兰公共许可证， 第2版
+
+2021年5月 http://license.coscl.org.cn/MulanPubL-2.0
+
+您对“贡献”的复制、使用、修改及分发受木兰公共许可证,第2版（以下简称“本许可证”）的如下条款的约束：
+
+0.   定义
+
+“贡献” 是指由“贡献者”许可在“本许可证”下的受版权法保护的作品，包括最初“贡献者”许可在“本许可证”下的作品及后续“贡献者”许可在“本许可证”下的“衍生作品”。
+
+“贡献者” 是指将受版权法保护的作品许可在“本许可证”下的自然人或“法人实体”。
+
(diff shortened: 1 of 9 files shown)


marktsuchida:
@Ryanyntc2013 Thank you for your pull request!

Would you be willing to use LGPL, BSD, MIT, or Apache 2.0 for the license? I'm asking because the Mulan PubL appears to be a strong copyleft license, which would conflict with distributing Micro-Manager builds that include proprietary code. For the same reason we also cannot accept GNU GPL.

Existing device adapters have the above-listed licenses that are permissive (or at most weakly copyleft, as with LGPL), GPL-compatible, Free (by FSF definition), and OSI-approved open source. Licenses would also ideally be on the [SPDX list](https://spdx.org/licenses/) so that we don't have to independently verify the free/open source status. We'd need to have a proper discussion if we are going to use a license that doesn't satisfy these criteria, because it may limit how we and others can use and (re)distribute Micro-Manager as a whole.

Also, I think we would like to separate out the files that belong to your SDK and put those in our private repository of dependencies. The directory in this repo should contain just the Micro-Manager device adapter code (`MMSCCam.{h,cpp}` if I understand right). This also means that your SDK files (`SCApi.h`, `SCDefines.h`, and `scsdk.lib`, if I understand correctly) can have a proprietary license if you like (but not Mulan PubL or GNU GPL), as long as it doesn't restrict the distribution of software built using them.

cxylh:
@marktsuchida

cxylh:
@marktsuchida Thank you for your reply. We are willing to use LGPL license. In addition, you can separate our sdk files (SCApi.h, SCDefines.h, and scsdk.lib). Do I need to resubmit a commit.

cxylh:
@marktsuchida 
Thank you for your reply. We are willing to use LGPL license. In addition, you can separate our SDK files (SCApi.h, SCDefines.h, and scsdk.lib). Do I need to resubmit a commit.

PR #538, opened 2024-12-16
add onImageSnapped, onSequenceAcquisitionStarted, onSequenceAcquisitionStopped events

tlambert03:
These events would make it easier for the application side to "know" that a new image buffer is available to display.

As with most callbacks here, I'm aware that this implementation is slightly leaky (in that some devices may snap/start/stop without these events being emitted).  That's certainly less than ideal, but it's still better than not knowing at all, and covers cases where the core API itself is used to snap/start/stop

### DIFF
--- a/MMCore/MMCore.cpp
+++ b/MMCore/MMCore.cpp
@@ -2543,6 +2543,12 @@ void CMMCore::snapImage() throw (CMMError)
             }
             waitForDevice(shutter);
          }
+
+         if (externalCallback_)
+         {
+            std::string cameraLabel = camera->GetLabel();
+            externalCallback_->onImageSnapped(cameraLabel.c_str());
+         }         
 		}catch( CMMError& e){
 			throw e;
 		}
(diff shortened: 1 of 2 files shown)


marktsuchida:
TODO:
- [ ] Check that MMStudio has no issues with MMCoreJ built with this (it shouldn't)
- [ ] Bump MMCore minor version at time of merge (not major since this should be source-compatible)

marktsuchida:
The "stopped" event needs to be also issued when the camera stops (because it finished the requested number of frames). In such cases the app may or may not call `stopSequenceAcquisition()`. Because that will happen on an arbitrary thread, we should make it the rule that it will always be asynchronous (but guaranteed to be called after the acquisition stopped, of course).

marktsuchida:
Ok to close this in favor of #659?

tlambert03:
lol, yes... sorry I had forgotten I already tried this before :)

PR #557, opened 2025-02-01
Core: Adds functions to get and set dxdz, dydz and optionalZ in pixel size conifgurations.

nicost:
See issues:
- https://github.com/micro-manager/micro-manager/issues/1984
- https://github.com/micro-manager/micro-manager/issues/2038

### DIFF
--- a/MMCore/MMCore.cpp
+++ b/MMCore/MMCore.cpp
@@ -5113,8 +5113,16 @@ void CMMCore::setPixelSizeAffine(const char* resolutionID, std::vector<double> a
 }
 
 /**
- * Sets the dx/dZ part of the angle between the scan axis and camera axies.
+ * Sets the angle between the camera's x axis and the axis (direction) 
+ * of the z drive.  This angle is dimensionless (i.e. the ratio of the 
+ * translation in x caused by a translation in z, i.e. dx / dz).  
+ * This angle can be different for different z drives (if there 
+ * are multiple Z drives in the system, please add the Core-Focus device
+ * to the pixel size configuration).  
  * See: https://github.com/micro-manager/micro-manager/issues/1984
+ *
+ * @param resolutionID   The pixel size configuration group name
+ * @param dxdz       Angle of the Z-stage axis with the camera axis (dimensionless)
  */
 void CMMCore::setPixelSizedxdz(const char* resolutionID, double dxdz)  throw (CMMError)
 {
@@ -5132,8 +5140,16 @@ void CMMCore::setPixelSizedxdz(const char* resolutionID, double dxdz)  throw (CM
 }
 
 /**
- * Sets the dy/dZ part of the angle between the scan axis and camera axies.
+ * Sets the angle between the camera's y axis and the axis (direction) 
+ * of the z drive.  This angle is dimensionless (i.e. the ratio of the 
+ * translation in y caused by a translation in z, i.e. dy / dz).  
+ * This angle can be different for different z drives (if there 
+ * are multiple Z drives in the system, please add the Core-Focus device
+ * to the pixel size configuration).  
  * See: https://github.com/micro-manager/micro-manager/issues/1984
+ *
+ * @param resolutionID   The pixel size configuration group name
+ * @param dydz       Angle of the Z-stage axis with the camera axis (dimensionless)
  */
 void CMMCore::setPixelSizedydz(const char* resolutionID, double dydz)  throw (CMMError)
 {
@@ -5154,8 +5170,11 @@ void CMMCore::setPixelSizedydz(const char* resolutionID, double dydz)  throw (CM
  * Sets the opimal Z stepSize (in microns).
  * There is no magic here, this number is provided by the person configuring the
  * microscope, to be used by the person using the microscope.
+ *
+ * @param resolutionID   The pixel size configuration group name
+ * @param optimalZ       Optimal z step in microns
  */
-void CMMCore::setPixelSizeOptimalZ(const char* resolutionID, double optimalZ)  throw (CMMError)
+void CMMCore::setPixelSizeOptimalZUm(const char* resolutionID, double optimalZ)  throw (CMMError)
 {
    CheckConfigPresetName(resolutionID);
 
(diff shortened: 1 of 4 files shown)


MMCore/MMCore.cpp
```diff
@@ -5112,7 +5112,65 @@ void CMMCore::setPixelSizeAffine(const char* resolutionID, std::vector<double> a
       std::fixed << std::setprecision(5) << affine[5];
 }
 
+/**
+ * Sets the dx/dZ part of the angle between the scan axis and camera axies.
+ * See: https://github.com/micro-manager/micro-manager/issues/1984
+ */
+void CMMCore::setPixelSizedxdz(const char* resolutionID, double dxdz)  throw (CMMError)
```
tlambert03, line 5119: just a comment about naming/documentation.  while i think anyone who's done some stage-scanning microscopy might look at `PixelSizedxdz` and suspect that perhaps this pertains to them, I think that the average API reader might be confused here, due to the overloaded usage of "Z" (here meaning "scan axis" as opposed to "optical axis").  

I think units need to be clearer ..

I see there was some prior discussion about this in (https://github.com/micro-manager/micro-manager/issues/1984#issuecomment-2195865832):

> > There is also the question of whether to record the angle or its tangent, i.e., the x-z skew ratio
> I have no preference here, as long as we define things so if the skew/angle is 0° then the metadata value is 0 and this is the default case for a "normal microscope."

the wording here suggests that dxdz should literally be the ratio of the image pixel size to the scan step size?  am I interpreting that correctly?  And if so, doesn't that suggest it's tightly tied to the value of `pixelSizeUm_` in a way that using the angle between the scan axis and the optical axis would not be?

marktsuchida: dx/dz is the ratio of the x-direction shift between two slices to the slice interval. So its units are um/um and it's not tied to the pixel size. But, yes, I think we need better documentation.

In the way I like to think, we should be recording the simplest piece of data that specifies how to correctly deskew a stack: (x shift parallel to the slice) per (z shift orthogonal to the slice). Hopefully this does not require specifying anything further about how the microscope was constructed -- computing dx/dz from hardware geometry and/or calibration should be the job of a method-specific tool. I'm assuming that this view is equivalent to @fadero's detailed definition in #1984 -- but please do correct me if you don't think so.

(And the same for y, of course.)

The "pixel size" in this function name (I assume) relates to the fact that the dx/dz setting applies to a pixel size preset (aka `resolutionID`). But it would be nice if there were a clearer name. Putting some word in front of of `dxdz` that indicates that this is about skewed scanning could perhaps help. Maybe `setPixelSizeDeskewDxDz` (can't say I love this either)?

fadero: > I'm assuming that this view is equivalent to @fadero's detailed definition in #1984 -- but please do correct me if you don't think so.

That's right!

nicost:
> Lets increment the MMCore minor version before merging.

Done

> Presumably dx/dz and dy/dz needs to be persisted in the config file. I'd be okay with adding this after it's been tested and we're absolutely sure about the data model (including the tieing to the pixel size group).

Interesting.  I am not quite sure where this happens now.  Only in the Java layer?  MMCore.cpp will read the pixel size configuration, but will not write it.  I guess that the function `saveSystemConfiguration` is not used anywhere?

> I think it would be nice if the doc comments could give a concise but complete definition of the values (the very long discussion thread that might confuse newcomers if it's the only reference).

I gave it a try.  Please provide suggestions for improvement.

> just a comment about naming/documentation. while i think anyone who's done some stage-scanning microscopy might look at PixelSizedxdz and suspect that perhaps this pertains to them, I think that the average API reader might be confused here, due to the overloaded usage of "Z" (here meaning "scan axis" as opposed to "optical axis").
> 
> I also think units need to be clearer in the docstrings ...

I am very open to better naming for these functions, but have no good ideas myself.  I added documentation, including dimensions.  Renamed the OptimalZ functions to include dimensions in the function name (which I like, certainly easier 
to read the code and know the dimensions).

> the wording here suggests that dxdz should literally be the ratio of the image pixel size to the scan step size? am I interpreting that correctly? And if so, doesn't that suggest it's tightly tied to the value of pixelSizeUm_ in a way that using the angle between the scan axis and the optical axis would not be?

i.e. does this ration change when inserting intermediate magnification (such as a tube lend; a Micro-Manager magnifier device will automatically adjust the pixel size returned by MMCore).  Good question!  I hope @fadero has insights!

fadero:
> the wording here suggests that dxdz should literally be the ratio of the image pixel size to the scan step size? am I interpreting that correctly? And if so, doesn't that suggest it's tightly tied to the value of pixelSizeUm_ in a way that using the angle between the scan axis and the optical axis would not be?

@tlambert03, you are correct that dx/dz (and dy/dz) are the ratio of the image pixel size to the scan step size. However, as @marktsuchida pointed out, dx/dz and dy/dz are dimensionless quantities (units cancel). Therefore, they're unrelated to absolute pixel size of the camera. Mathematically, specifying the ratio of two vectors is equivalent to specifying the angle. 

Under "normal" circumstances (i.e. a z-piezo stage whose translation axis is perfectly orthogonal to both the camera's x and y axes), dx/dz and dy/dz are 0. This means that a translation along z produces a translation along x of x_translation = dx/dz * z_translation. Units only come into play for the pixel size when you bring in units from the z stage.

tlambert03:
Yep, all sounds good.  Thanks for confirming!

marktsuchida:
> > Presumably dx/dz and dy/dz needs to be persisted in the config file. I'd be okay with adding this after it's been tested and we're absolutely sure about the data model (including the tieing to the pixel size group).
> 
> Interesting. I am not quite sure where this happens now. Only in the Java layer? MMCore.cpp will read the pixel size configuration, but will not write it. I guess that the function `saveSystemConfiguration` is not used anywhere?

Yes, as things currently stand the saving code that is actually used is in MMStudio but `CMMCore::loadSystemConfiguration()` is used for loading. So MMCore will need to first support reading the new config commands; then MMStudio (and pymmcore-plus) can be updated to actually save the values.

PR #562, opened 2025-02-06
ASI Stage: add property caching

nicost:
On some older microscope systems, I noted that much of the lag in the software is caused by the System updating its cache, which caused the core to ask each device for the value of all of its properties.  The reason for doing is that properties can be changed as a side effect of changing another property, or, more often, there is another interface to the device (such as buttons or a joystick) that also can change these properties.  

Nevertheless, there are also many properties that can be efficiently cached within the device adapter.  If only the device adapter can change these values, and if they do not change as a side effect of changing some other property of the device, then it is fine to read the value of the property on startup and use that cached value (updating the cached value whenever the property is changed by MMCore).  

This PR caches the Version of the controller (which certainly never changes) and also the Acceleration property of the XY Stage.  If this looks good, I am happy to use the same procedure with other properties (XY stage position and Z stage position are obviously not to be cached, are there other properties that should always be read from the device)?  Let me know if this looks like the right approach!

### DIFF
--- a/DeviceAdapters/ASIStage/ASIBase.cpp
+++ b/DeviceAdapters/ASIStage/ASIBase.cpp
@@ -14,6 +14,7 @@ ASIBase::ASIBase(MM::Device* device, const char* prefix) :
 	initialized_(false),
 	device_(device),
 	oldstagePrefix_(prefix),
+	version_("undefined"),
 	port_("Undefined")
 {
 	versionData_ = VersionData();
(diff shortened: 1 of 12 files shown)


marktsuchida:
It looks logically correct and non-problematic for these particular properties.
Cc: @bls337

bls337:
Thank you for your effort Nico! @nicost 

Feel free to merge the part where you cache the version string.

For the XY acceleration and other properties could we try another approach?

I can try a few things out on Monday morning or if you're feeling adventurous you can use ASITiger as a reference.

I'd like to try an approach similar to how we do things on Tiger:

```
if (!refreshProps_ && initialized_)
	return DEVICE_OK;
```

I think the XYStage and ZStage are fairly safe for caching, CRISP may be a bit more difficult.

nicost:
Too late, sorry! Just finished going through all the XY stage and Z stage properties. Would love it if you would take a crack at the others.  I don't think the approach matters that much.  I'll test the current code (hopefully tomorrow) and merge if it works.  Not worth your effort to do the same thing with a different approach, but again, it would be awesome if you could look at the other devices.

bls337:
@nicost No worries, thank you! We appreciate the effort, I will use the same approach with CRISP and the other devices. Please feel free to merge after you test the code.

PR #575, opened 2025-02-25
MMDevice: Depercate unsafe overloads of InsertImage()

marktsuchida:
In MMDevice we avoid passing nontrivial C++ objects because they are not always compatible between different types of builds (e.g., MMCore built in Release mode and device adapter built in Debug mode using MSVC). A couple of functions have been violating this rule by passing `ImgBuffer` or `Metadata`. Deprecate these.

Newly deprecate `InsertImage(const Device*, const ImgBuffer&)`.

Newly deprecate `InsertImage(const Device*, const unsigned char*, unsigned, unsigned, unsigned, const Metadata*, bool)`.

Add deprecation warnings for the newly and previously deprecated overloads.

However, **undeprecate** `InsertImage(const Device*, const unsigned char*, unsigned, unsigned, unsigned, const char*, bool)` because it is widely used and reasonable for monochrome-only cameras.

This leaves just 2 overloads of `InsertImage()` undeprecated, one with `nComponents` and one without.

@henrypinkard I don't think this conflicts with #570 (which changes the implementations of these functions), but let me know if you have any concerns.

### DIFF
--- a/DeviceAdapters/PICAM/PICAMUniversal.cpp
+++ b/DeviceAdapters/PICAM/PICAMUniversal.cpp
@@ -2556,24 +2556,22 @@ int Universal::PushImage(const unsigned char* pixBuffer, Metadata* pMd )
    int nRet = DEVICE_ERR;
    MM::Core* pCore = GetCoreCallback();
    // This method inserts a new image into the circular buffer (residing in MMCore)
-   nRet = pCore->InsertMultiChannel(this,
+   nRet = pCore->InsertImage(this,
          pixBuffer,
-         1,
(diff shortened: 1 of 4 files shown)


henrypinkard:
No concerns, that makes sense to me

henrypinkard:
Probably would be good to mark these as deprecated in corecallback.h as well

marktsuchida:
> Probably would be good to mark these as deprecated in corecallback.h as well

#589. It shouldn't be necessary to use the compiler directives here because only MMCore internal code has access to CoreCallback directly.

Issue #258, opened 2022-10-16
Standardized device properties


henrypinkard:
Based on discussions @marktsuchida and I have had and related to the camera API upgrade (https://github.com/micro-manager/mmCoreAndDevices/issues/243). 

We need to come up with a better mechanism for standardized device properties. This issue to is to solicit ideas about how best to do this.

At a high level this is because: 

Properties are very useful for capturing certain types of data and functionality, and it creates unnecessary work and the potential to introduce errors when you have to recreate this functionality for property-like device API calls. For example, creating methods `GetX()`, `HasX()`, `GetXMaxValue()` etc. for each `X` when all this would already exists if making `X` a property.

However, simply using a property for a required feature of the API doesn't work well as present. It is not clear to device adapter authors or high level code users that certain properties are required as opposed to device specific. `Exposure` and `Binning` properties are both required, but inconsistently implemented in part as a result of these shortcomings (see https://github.com/micro-manager/mmCoreAndDevices/issues/38 for more info about binning and how to fix it).

`Exposure` and `Binning` have both properties and API calls. This seems likely to have resulted from the fact that functions in MM control the generic API for devices, while properties (usually) enable functionality unique to specific devices. This duplicate functionality may have been present so that exposure could be included in config groups, to assist with switching channels etc.

The camera API has also made apparent some shortcomings of the property system. Several features related to camera triggering will be implemented following the GenICam specification as closely as possible. The most natural way to do this will be through standardized properties, and on many cameras this will create 50-70 new properties because of all the combinations of trigger types and settings. For example:

`FrameStartTriggerMode`: `On`
`FrameStartTriggerSource`: `Hardware`
`FrameStartTriggerDelay`: 5
`ExposureEndTriggerMode`: `Off`
etc...

## What needs to be added to properties?
**A mechanism to group logically related properties**. Ideally in some kind of hierarchy. For example: `Triggers//FrameStart//Mode`: `On`. A simple way of achieving this could be to designate a special character or sequence or characters to denote a new level in the hierarchy (e.g. `//`). This would allow UIs to group properties in a way that would be much easier to work with.

**Mechanisms and conventions to discriminate:** 
- **Required properties** with a standard name (and maybe standard values) that every device of a certain type must implement. E.g. `exposure` for cameras. Implementing this may involve creating `virtual` properties in `MMCore` or `MMDevice` to throw compile-time errors for missing requirements. 
[...]

PR #584, opened 2025-02-28
Standard properties for device adapters

henrypinkard:
# Standard Properties for Device Adapters

This PR introduces a standard properties system to Micro-Manager device adapters, addressing #258 

### Motivation
Currently, standardized functionality across devices is only expressed through methods in different device types. However, many features are more naturally expressed as properties (eg camera triggering in the new camera API #243). Furthermore, many device adapters have common functionality that is accessed through properties with identical or nearly identical names (e.g. Gain in a camera). This PR allows certain properties to be made part of the API.

To the higher-level application, standard properties function just like regular device properties, only with an `api//` prefix, signaling that they have consistent meaning across different device adapters. This is thus for now a fully backwards-compatible, opt-in mechanism. Regular properties without the `api//` prefix will continue to work indefinitely

For device adapter developers, each standard property has a dedicated creation method (e.g., `CreateTriggerTypeStandardProperty`) that can only be called in device types to which it is applicable

Two dummy test standard properties are implemented in this PR, added to the demo camera

### How to Create a New Standard Property

1. Add specification of the standard property in MMDevice.h:

```cpp
static const std::vector<std::string> allowedValues = {"Internal", "External"};
static const MM::StandardProperty g_TriggerTypeStandardProperty(
   "TriggerType",                // name
   String,                   // type
   false,                    // isReadOnly
   false,                    // isPreInit
   allowedValues,                       // allowedValues 
   {},                       // requiredValues (empty vector)
   PropertyLimitUndefined,   // lowerLimit
   PropertyLimitUndefined,   // upperLimit
   false                     // required
);
```

2. Link it with a particular device type using this macro in MMDevice.h:

```cpp
LINK_STANDARD_PROP_TO_DEVICE_TYPE(CameraDevice, g_TriggerTypeStandardProperty)
```

3. Add a dedicated creation method in CDeviceBase:

```cpp
int CreateTriggerTypeStandardProperty(const char* value, MM::ActionFunctor* pAct = 0)
```

### Technical notes/questions
@marktsuchida 
- Compile-time checking of standard property methods using SFINAE patterns ensures they can only be used by appropriate device types. The linking macro enables the same standard properties to be used across multiple device types while throwing compile-time errors if they're used with incompatible device types. In order to implement these compile-time checks, I had to move the ``DeviceType`` from the class definition to the declaration and make it a `constexpr`. These was a note about compiler warnings related to this in `MMDevice.cpp`, but these didn't seems to happen. Perhaps because of the use of `constexpr`? or perhaps because this comment is 11 years old and no longer applicable). @marktsuchida you think this is okay?
- Runtime checks of required values are performed in the initialization of device instances in the core. It would have been nice to do this directly in the device layer, but I didn't see any way to do this given that there was no common superclass initialization called


marktsuchida:
> * Compile-time checking of standard property methods using SFINAE patterns ensures they can only be used by appropriate device types. The linking macro enables the same standard properties to be used across multiple device types while throwing compile-time errors if they're used with incompatible device types. In order to implement these compile-time checks, I had to move the `DeviceType` from the class definition to the declaration and make it a `constexpr`. These was a note about compiler warnings related to this in `MMDevice.cpp`, but these didn't seems to happen. Perhaps because of the use of `constexpr`? or perhaps because this comment is 11 years old and no longer applicable). @marktsuchida you think this is okay?

Those comments described the requirement prior to C++11, so yes, it's fine.

> * Runtime checks of required values are performed in the initialization of device instances in the core. It would have been nice to do this directly in the device layer, but I didn't see any way to do this given that there was no common superclass initialization called

I think it is fine to do it in MMCore -- in fact it gives me more confidence because it can't be messed with by the device adapter.

MMDevice/MMDevice.h
```diff
@@ -205,6 +208,104 @@ namespace MM {
...
+
+   // The top struct enables compile time checking of standard property
+   // creation methods (ie that you can call the CreateXXXXStandardProperty
+   // in a device type that doesn't support it)
+   // The bottom part adds the standard property to a runtime registry
+   // which enables higher level code the query the standard properties
+   // associated with a particular device type and check that everything
+   // is correct (ie all required properties are implemented after
+   // device initialization)
+   #define LINK_STANDARD_PROP_TO_DEVICE_TYPE(DeviceType, PropertyRef) \
```
marktsuchida, line 261: Since macros are not confined to a namespace, best to name with a prefix. Here, perhaps `MM_INTERNAL_` so that it's clear that user device code should not use this macro.

MMDevice/MMDevice.h
```diff
@@ -205,6 +208,104 @@ namespace MM {
    };
 
 
+   ////////////////////////////
+   ///// Standard properties
+   ////////////////////////////
```
marktsuchida, line 213: Much of the machinery here (the parts that are not intended for use by user device code) can be placed in namespace `MM::internal` to make this clear.

MMDevice/Property.h
```diff
@@ -27,9 +27,64 @@
...
 namespace MM {
 
+// Standard Properties
+const char* const g_KeywordStandardPropertyPrefix = "api//";
+
+// Define NaN for use in property definitions
+const double PropertyLimitUndefined = std::numeric_limits<double>::quiet_NaN();
+
+// Standard property metadata structure
+class StandardProperty {
```
marktsuchida, line 41: I think this can be a plain `struct` with all public fields. It clarifies that this is a pure-data object. There are no checks on the constructor arguments currently and `CreateStandardProperty` can perform checks when/if necessary.

Also eventually in C++20 we'll be able to use [designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers).

marktsuchida (commented):
Some initial comments.

I think the major issue that needs to be solved is how we support future migration upon adding new standard properties. If we have a YES/NO distinction between implementing or not implementing all required standard properties, it will be very difficult to ever add any new required ones; this then calls into question the point of having required standard properties in the first place.

There may certainly be some properties that require others to also exist, but that kind of complex check is probably best limited to runtime.

henrypinkard:
> I think the major issue that needs to be solved is how we support future migration upon adding new standard properties. If we have a YES/NO distinction between implementing or not implementing all required standard properties, it will be very difficult to ever add any new required ones; this then calls into question the point of having required standard properties in the first place.

I view the requirement as useful for when making new device types/APIs. I think that using the `CCameraBase` etc classes, we can add default implementations of standard properties. When making new base classes (either for the same device type, like your suggestion the camera API, or entirely new types of devices, we can ensure that devices developers have implemented all required properties. So in general a way to enforce that all devices of a particular type implement required properties going forward is to mark the current base class legacy, and add standard props to the device type that are unimplemented in the new base class. 

No need to merge this now. I will use it when developing the camera API and see if I get further insights about how best to structure it

henrypinkard:
In implementing the new camera API, I've gained more clarity on how to effectively implement standard properties:

I've removed the required field entirely. All standard properties for a device type now default to being required, but this requirement can be bypassed by calling a `SkipXXXXStandardProperty` method in the device adapter. Failing to either implement or skip causes an initialization error. This requirement can be managed at the base class level (e.g., in `CCameraBase`).

The design allows declaring a set of standard properties for a device type (e.g., `MM::Camera`) with different behavior across subclasses. For example:

- The old camera API base class automatically skips all standard properties, preserving backward compatibility while allowing specific adapters to opt-in if desired
- The new camera API base class does not skip the standard properties, requiring adapter developers to either implement them or explicitly skip them

This approach accomplishes several goals:

- Preserves backward compatibility for existing devices
- Ensures developers using newer APIs are aware of standard properties while maintaining flexibility if a particular device doesn't implement them
- Provides a migration path where both standard and non-standard properties can coexist. In many cases, moving to a standard property may just require adding an existing property callback to the standard property creation function. The old non-standard and new standard properties (which control the same functionality) can then both exist
- Simplifies the camera API by implementing GenICam conventions without requiring numerous getter/setter methods

For application layer code, checking for standard property support is straightforward: `core.hasProperty(device, 'api//prop_name')`

Finally, I added the ability to delete a standard property, because I noticed on the Basler camera that the existence of properties often depended on the values of others. 

Nothing else to do on this on my end. Ready to merge, or happy to make further refinements if needed.

PR #588, opened 2025-03-04
Automatically clear circular buffer in continuous sequence acquisition

henrypinkard:
This adds an overwrite mode to the circular buffer, which means that when an image is inserted and there's no more space, it will automatically clear itself. Currently, most cameras do this manually when there is an error on insert. 

This is motivated by the need to deprecate the ability of devices to clear and allocate the buffer in the core. This should instead be done by higher level application code.

One thing I don't fully understand is the `stopOnOverflow` argument of `startSequenceAcquisition`. Some, but not all, cameras appear to have logic that changes behavior based on this flag, but I'm not aware of any higher level code that ever sets this to `false` when running a finite sequence of N frames. 

Regardless, `stopOnOverflow` shouldn't be an issue for merging this, because the only time we set `overwrite=true` is when calling `startContinuousSequenceAcquisition`, where `stopOnOverflow` is always false

### DIFF
--- a/MMCore/CoreCallback.h
+++ b/MMCore/CoreCallback.h
@@ -93,7 +93,7 @@ public:
    /*Deprecated*/ int InsertMultiChannel(const MM::Device* caller, const unsigned char* buf, unsigned numChannels, unsigned width, unsigned height, unsigned byteDepth, Metadata* pMd = 0);
   
    /*Deprecated*/ void ClearImageBuffer(const MM::Device* caller);
-   /*Deprecated*/ bool InitializeImageBuffer(unsigned channels, unsigned slices, unsigned int w, unsigned int h, unsigned int pixDepth);
+   bool InitializeImageBuffer(unsigned channels, unsigned slices, unsigned int w, unsigned int h, unsigned int pixDepth);
 
    int AcqFinished(const MM::Device* caller, int statusCode);
    int PrepareForAcq(const MM::Device* caller);
--- a/MMCore/CoreCallback.h
+++ b/MMCore/CoreCallback.h
@@ -91,8 +91,9 @@ public:
    /*Deprecated*/ int InsertImage(const MM::Device* caller, const unsigned char* buf, unsigned width, unsigned height, unsigned byteDepth, unsigned nComponents, const Metadata* pMd = 0, const bool doProcess = true);
 
    /*Deprecated*/ int InsertMultiChannel(const MM::Device* caller, const unsigned char* buf, unsigned numChannels, unsigned width, unsigned height, unsigned byteDepth, Metadata* pMd = 0);
-   void ClearImageBuffer(const MM::Device* caller);
-   bool InitializeImageBuffer(unsigned channels, unsigned slices, unsigned int w, unsigned int h, unsigned int pixDepth);
+  
+   /*Deprecated*/ void ClearImageBuffer(const MM::Device* caller);
+   /*Deprecated*/ bool InitializeImageBuffer(unsigned channels, unsigned slices, unsigned int w, unsigned int h, unsigned int pixDepth);
 
    int AcqFinished(const MM::Device* caller, int statusCode);
    int PrepareForAcq(const MM::Device* caller);
--- a/MMCore/MMCore.cpp
+++ b/MMCore/MMCore.cpp
@@ -2878,7 +2878,7 @@ void CMMCore::startSequenceAcquisition(const char* label, long numImages, double
       throw CMMError(getCoreErrorText(MMERR_CircularBufferFailedToInitialize).c_str(), MMERR_CircularBufferFailedToInitialize);
    }
    cbuf_->Clear();
-   cbuf_->SetOverwriteData(false);
+   cbuf_->SetOverwriteData(!stopOnOverflow);
    LOG_DEBUG(coreLogger_) <<
       "Will start sequence acquisition from camera " << label;
    int nRet = pCam->StartSequenceAcquisition(numImages, intervalMs, stopOnOverflow);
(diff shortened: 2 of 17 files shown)


MMCore/MMCore.cpp
```diff
@@ -2833,6 +2833,7 @@ void CMMCore::startSequenceAcquisition(long numImages, double intervalMs, bool s
 				throw CMMError(getCoreErrorText(MMERR_CircularBufferFailedToInitialize).c_str(), MMERR_CircularBufferFailedToInitialize);
 			}
 			cbuf_->Clear();
+         cbuf_->SetOverwriteData(false);
```
marktsuchida, line 2836: This should be `cbuf_->SetOverwriteData(!stopOnOverflow)`.

The other two places are correct (we default to not overwriting, except for "continuous" sequence acquisition).

MMDevice/MMDevice.h
```diff
@@ -1423,8 +1423,8 @@ namespace MM {
        */
       virtual int InsertImage(const Device* caller, const unsigned char* buf, unsigned width, unsigned height, unsigned byteDepth, const char* serializedMetadata, const bool doProcess = true) = 0;
 
-      virtual void ClearImageBuffer(const Device* caller) = 0;
-      virtual bool InitializeImageBuffer(unsigned channels, unsigned slices, unsigned int w, unsigned int h, unsigned int pixDepth) = 0;
+      MM_DEPRECATED(virtual void ClearImageBuffer(const Device* caller)) = 0;
+      MM_DEPRECATED(virtual bool InitializeImageBuffer(unsigned channels, unsigned slices, unsigned int w, unsigned int h, unsigned int pixDepth)) = 0;
```
marktsuchida, line 1427: `InitializeImageBuffer` serves a purpose that is unrelated to Live mode "continue by overwriting", so should not be deprecated in this PR.

It exists to allow cameras to trigger the memory allocation (which can take time) before opening the shutter, in order to minimize unnecessary exposure. Cameras that don't call it in `StartSequenceAcquisition()` are technically buggy.

MMCore/CoreCallback.h
```diff
@@ -91,8 +91,9 @@ class CoreCallback : public MM::Core
    /*Deprecated*/ int InsertImage(const MM::Device* caller, const unsigned char* buf, unsigned width, unsigned height, unsigned byteDepth, unsigned nComponents, const Metadata* pMd = 0, const bool doProcess = true);
 
    /*Deprecated*/ int InsertMultiChannel(const MM::Device* caller, const unsigned char* buf, unsigned numChannels, unsigned width, unsigned height, unsigned byteDepth, Metadata* pMd = 0);
-   void ClearImageBuffer(const MM::Device* caller);
-   bool InitializeImageBuffer(unsigned channels, unsigned slices, unsigned int w, unsigned int h, unsigned int pixDepth);
+  
+   /*Deprecated*/ void ClearImageBuffer(const MM::Device* caller);
+   /*Deprecated*/ bool InitializeImageBuffer(unsigned channels, unsigned slices, unsigned int w, unsigned int h, unsigned int pixDepth);
```
marktsuchida, line 96: (See comment in MMDevice.h.)

marktsuchida (changes requested):
Thanks, @henrypinkard. It's so much easier to confidently review a narrowly targeted PR like this!

MMCore/MMCore.cpp
henrypinkard: Why would the other instance of `startSequenceAcquisition` not need this same change?

henrypinkard: Or perhaps did you mean that that the two `startSequenceAcquisition` functions should be `SetOverwriteData(!stopOnOverflow)`, while `startContinuousSequenceAcquisition` should be `SetOverwriteData(true)`?

marktsuchida: I meant that this particular line needs the change to `!stopOnOverflow`; the other two places where you call `SetOverwriteData` are already correct and don't need any change.

henrypinkard: I'm still confused by this. So you're saying that `startSequenceAcquisition(long numImages, double intervalMs, bool stopOnOverflow)` should be `!stopOnOverflow` but `startSequenceAcquisition(const char* label, long numImages, double intervalMs, bool stopOnOverflow)` can remain `false`? Why wouldn't these be the same? Isn't the difference between the methods just that you're providing the camera name in one case but using the core camera in the other? Am I missing something?

marktsuchida: Ah, sorry, my bad. You're right -- both should honor `!stopOnOverflow`. I probably missed that the second overload of `startSequenceAcquisition` had the `stopOnOverflow` parameter.

`startContinuousSequenceAcquisition` is tricky. It is _intended_ for Live view, which would be okay with overwriting, but the method name and documentation don't suggest this. However, `CCameraBase` implements the single-argument overload of `StartSequenceAcquisition` (called only by `startContinuousSequenceAcquisition`) by passing `false` for `stopOnOverflow`, so the actual behavior has always been to allow overwriting. The documentation could use a fix.

henrypinkard: Got it, thanks for clarifying!

henrypinkard:
@marktsuchida ready to merge?

marktsuchida:
Code-wise, this is ready to merge, but let's wait a bit for a day on which other major changes are not going into the nightly build (we upgraded to Java 11 today).

- [ ] Bump MMCore minor version
- [ ] Bump MMDevice interface version

The device interface version needs to be bumped, despite the apparent lack of change, because there is a behavioral change: camera adapters written assuming this change will fail Live when the buffer fills up if run with an older Core. We may want to coordinate this version increment with #640 and the actual removal of `ClearImageBuffer()`.

PR #597, opened 2025-03-14
Add realistic (non-instantaneous) movement to DemoXYStage position, and expose velocity as property

tlambert03:
For testing, purposes, it would be nice if the XY demo stage didn't immediately return the new commanded stage position.  This PR makes it so that calling `GetPositionSteps` returns a realistic value based on the velocity and the time since the last call to setPositionSteps...  It also exposes velocity as a property.

@marktsuchida, please let me know what you think... haven't compiled and tested yet

### DIFF
--- a/DeviceAdapters/DemoCamera/DemoCamera.cpp
+++ b/DeviceAdapters/DemoCamera/DemoCamera.cpp
@@ -3594,7 +3594,7 @@ posX_um_(0.0),
 posY_um_(0.0),
 busy_(false),
 timeOutTimer_(0),
-velocity_(10.0), // in micron per second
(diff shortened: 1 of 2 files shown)


marktsuchida:
This was bothering me in a few scenarios recently, so wonderful. It compiles. I pushed a comment change (I almost got confused there -- all our units are um and ms, so you had it right).

For testing stage control GUIs, I'd also recommend [NotificationTester](https://micro-manager.org/NotificationTester), which has the additional ability to simulate the various ways in which motion devices are actually implemented.

tlambert03:
brilliant.  thanks!

PR #603, opened 2025-03-18
Clarify required/optional functions of pumps

marktsuchida:
Another followup to #462.

I think the required/optional distinction got partly lost when the pressure and volumetric pumps were separated. @Lars-Kool Do the changes proposed here correctly capture your intent? Let me know if changes are needed.

(The DeviceBase classes should leave out implementations for required functions, so that device code will get a compile error if the function is not implemented.)


Lars-Kool:
@marktsuchida, indeed, these changes are as I intended. I most likely forgot to update the comments after the pressure and volumetric pump split. I should have been more careful.

PR #605, opened 2025-03-24
IDS device adapter

Lars-Kool:
Hi @nicost and @marktsuchida,

I've updated the device adapter for IDS cameras. It now makes certain features optional (like analog/digital gain and auto whitebalance), such that cameras that don't support these features no longer crash.

Best,
Lars


marktsuchida:
Thanks, @Lars-Kool!

PR #616, opened 2025-04-09
WPI syringe pump adapter

Lars-Kool:
It's me again. I can finally start sharing the device adapter I've had running locally. This one is for WPI (World Precision Instruments) syringe pumps.


marktsuchida:
Thanks, @Lars-Kool!

I think it will be good to use a more specific name than `WPI` or `WorldPrecisionInstruments`, just because they have quite a few types of products and device adapters for other ones may be added in the future. Do you think `WPI_Aladdin` would be a good name? Or would something like `WPI_pumps` be better (if you think this may extend beyond just the Aladdin series)?

I pushed some commits to:
- Avoid `using namespace std` -- this tends to come back and [bite](https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice) us later
- Clean up some `std::string` usage
- Tentatively rename the device adapter to `WPI_Aladdin` (happy to change)

Other than the naming issue, I think this is good to merge.

I'll resolve the `micromanager.sln` conflict when merging to avoid double-handling.

Lars-Kool:
Hi @marktsuchida,

Renaming the device adapter to ``WPI_Aladdin`` works for me. It really depends on how you envision the Hub being used. Is it to only initialize all WPI (Aladdin) pumps, or can we repurpose it to initialize all WPI devices (like how the Demo Hub initializes not one specific type of device, but all demo devices).

There are arguments to be made to not crowd the list of available device adapters, as it can be hard to find the thing you want. But, there are also arguments to not hide functionality behind a generic name. Personally, I would create one hub per manufacturer, and then have the hub initialize all devices of that manufacturer. But, I leave that decision completely up to you.

PR #626, opened 2025-05-05
Switch Windows build to VS2022 (v143) toolset

marktsuchida:
This includes the required fixes for TIScam, MeadowlarkLC, and DahengGalaxy.


marktsuchida:
- [x] Confirm that the vcredist we ship with the Windows installer is at least as new as the VS2022 we use to build

marktsuchida:
- [ ] Ant script (in micro-manager/micro-manager) needs to require VS2022

marktsuchida:
- [ ] MMStudio crashes if `MMCoreJ_wrap.dll` is built with VS2022 and running with the currently-shipped JRE (Temurin 1.8.0_152)

Rescued by an `MMCoreJ_wrap.dll` (only) built with VS2019, or the latest Java 8 JRE (Temurin 1.8.0_452). So we need to upgrade our JRE first.

marktsuchida:
This should be good to merge after merging https://github.com/micro-manager/micro-manager/pull/2100, but let's not do it today, in the interest of not changing more than one major thing in the same nightly build. (We already merged https://github.com/micro-manager/micro-manager/pull/2101 today).

PR #627, opened 2025-05-05
feat: allow demo xy stage to respond to Stop()

tlambert03:
currently, stopping the XY demo stage doesn't stop it's virtual motion.

### DIFF
--- a/DeviceAdapters/DemoCamera/DemoCamera.cpp
+++ b/DeviceAdapters/DemoCamera/DemoCamera.cpp
@@ -3592,7 +3592,6 @@ CXYStageBase<CDemoXYStage>(),
 stepSize_um_(0.015),
 posX_um_(0.0),
 posY_um_(0.0),
-busy_(false),
 timeOutTimer_(0),
 velocity_(10.0), // in mm per second (= um/ms)
 initialized_(false),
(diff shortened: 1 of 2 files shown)


DeviceAdapters/DemoCamera/DemoCamera.cpp
```diff
@@ -3771,6 +3770,31 @@ void CDemoXYStage::ComputeIntermediatePosition(
...
+      ComputeIntermediatePosition(now, posX_um_, posY_um_);
+   }
+   // No active motion → posX/Y already hold the last settled values
+
+   // Drop the timer so Busy() instantly goes idle
+   delete timeOutTimer_;
+   timeOutTimer_ = nullptr;
+
+   // Core listeners expect a position‑changed notification
+   (void)OnXYStagePositionChanged(posX_um_, posY_um_);
```
marktsuchida, line 3787: Maybe better to avoid emitting this event if already stopped? Not that it is technically wrong.

DeviceAdapters/DemoCamera/DemoCamera.cpp
```diff
@@ -3771,6 +3770,31 @@ void CDemoXYStage::ComputeIntermediatePosition(
...
+}
+
+int CDemoXYStage::Stop()
+{
+   MMThreadGuard g(this->stopLock_);
```
marktsuchida, line 3792: Do we need this lock? Note that calls into device adapters are already protected by a per-adapter lock.

tlambert03: aha!  ok, I was (mindlessly) copying patterns I saw elsewhere in here.  will remove

DeviceAdapters/DemoCamera/DemoCamera.cpp
tlambert03: have another look

PR #634, opened 2025-05-14
MMCore: Keep channel group choices up to date

### LINKED ISSUE #631: updateAllowedChannelGroups not called by defineConfig
found this by manually reimplementing a loadSystemConfiguration method in python, and when I hit the `setConfig("System", "Startup")` call, the core was unaware of my config group called "Channels" ... even though I had previously defined it using `defineConfig(....)` (but *not* by first declaring an empty config using `defineConfigGroup` ... which would have avoided the issue, since that method *does* call `updateAllowedChannelGroups`)

In the C++ code, this is prevented by manually calling `updateAllowedChannelGroups` [here](https://github.com/micro-manager/mmCoreAndDevices/blob/9dcb3ffc1ca83565759a8f7b2f214c76a6581c0d/MMCore/MMCore.cpp#L7597) ... but that seems like a flaky fix

marktsuchida:
They were not correctly updated when

- A config group is created via `defineConfig()` without a call to `defineConfigGroup()`
- All config groups are removed by `unloadAllDevices()`

Also cleaned up some related code.

Closes #631.

### DIFF
--- a/MMCore/ConfigGroup.h
+++ b/MMCore/ConfigGroup.h
@@ -217,11 +217,7 @@ public:
     */
    bool isDefined(const char* groupName)
    {
-      std::map<std::string, ConfigGroup>::iterator it = groups_.find(groupName);
-      if (it == groups_.end())
-         return false;
-      else
-         return true;
+      return groups_.find(groupName) != groups_.end();
    }
 
    /**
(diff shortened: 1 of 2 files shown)


marktsuchida:
Will test tomorrow because I spent my evening [fixing](https://developercommunity.visualstudio.com/t/VS-1714-incorrectly-forms-INCLUDE-env/10904181) Visual Studio....

PR #636, opened 2025-05-15
Support 28800 baudrate

marktsuchida:
This adds 28800 as an option for SerialManager baudrate, and adds it to the port scanning for ASIStage.

@bls337 This is motivated by your code comment in #635. I assume this should be good but let me know.


bls337:
Thank you @marktsuchida, this looks great!

PR #645, opened 2025-05-20
Fixed commands in CoherentOBIS DeviceAdapter.

ajbutler195:
Addressing issues described in the following image.sc post: https://forum.image.sc/t/i-cannot-control-coherent-obis-laser-properly-with-micromanager/70937/8.

Using the CoherentOBIS LS/LX laser with micro-manager, the device reports a state of "2" (invalid) and repeatedly reverts to this state. With some testing and packet sniffing, it appears some commands that succeed in PuTTY-- with SYST: or SOUR:-- are sent to the device as SYST1: and SOUR1, causing the device errors and constant switching of states as the device can't be queried.

In the DeviceAdapter, these are hard-coded (SYST1 and SOUR1, with a strange comment "Initialize laser??"), not some enumeration. Fixing the DeviceAdapter and rebuilding with Device Interface 71 (the version our build is using) resolved our errors.

### DIFF
--- a/DeviceAdapters/CoherentOBIS/CoherentOBIS.cpp
+++ b/DeviceAdapters/CoherentOBIS/CoherentOBIS.cpp
@@ -25,7 +25,6 @@
 #ifdef WIN32
    #include <windows.h>
 #endif
-#include "FixSnprintf.h"
 
 #include "CoherentOBIS.h"
 
--- a/DeviceAdapters/CoherentOBIS/CoherentOBIS.cpp
+++ b/DeviceAdapters/CoherentOBIS/CoherentOBIS.cpp
@@ -25,6 +25,7 @@
 #ifdef WIN32
    #include <windows.h>
 #endif
+#include "FixSnprintf.h"
 
 #include "CoherentOBIS.h"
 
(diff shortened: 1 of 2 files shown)


marktsuchida:
@ajbutler195 Thank you for the fix!

It would be good to make sure that this change doesn't break the code for other (older? different?) models before we merge this.

Section 9.2.5 "Device Selection Syntax" of the [manual](https://www.coherent.com/resources/manuals/lasers/obis-lx-ls-operators-manual.pdf) says that `SYST` is equivalent to `SYST0`, and the numeric suffix selects the device. It also says that `0` is the OBIS remote and `1` is the laser, although there can also be multi-laser configurations. But when the laser is directly connected to the host computer, `0` is the laser.

I think this explains why the code was written with `1` hard-coded, while you need `0` (or nothing). But presumably other people will need it to be `1`.

So the best way to fix this is probably to add a pre-initialization property (maybe call it `DeviceIndex`), which defaults to `1` (so as not to break the configuration for existing users who have it working that way), but allows you to also select `0`.

Would you be interested in adding that? If not, I'm also happy to draft the change if you can help with testing it (I don't think I have an OBIS to test with at the moment).

ajbutler195:
@marktsuchida can do! I have a laser I can test with for now. Will push some additional commits for review.

ajbutler195:
I've tested for the case in which deviceIndex_ is set pre-initialization to 0 in the configuration file, with success as expected. Switching the deviceIndex_ value to 1 produced a failure (as expected), though commands were still sent with proper formatting (SYST1, while my instrument configuration expects SYST0). I'm not sure if I can test for success in this case without a remote.

marktsuchida:
@ajbutler195 Thanks, this looks good (as of afe6d93). If you're done, we can merge.

PR #646, opened 2025-05-21
MMCore: Disable callback notifications in dtor

marktsuchida:
This prevents various issues because arbitrary code in destructors is never a good idea. Also helps with keeping things safe for Python bindings.


tlambert03 (approved):
Thanks

PR #648, opened 2025-05-22
Mock device adapters for MMCore unit testing

marktsuchida:
One of the main reasons we haven't added more unit tests for MMCore has been that most things cannot be tested without loading devices, but using any given device adapter (whether DemoCamera or SequenceTester) is too indirect, inflexible, and fragile.

This PR adds the ability to create ad-hoc, purpose-crafted mock device adapters in C++ unit tests. Tests can directly interact with such self-created mocks to ensure that MMCore is interacting with devices as expected.

There might be future extensions of this that allow testing via (Python/Java) language bindings, etc., but for now, it is a minimum implementation intended only for unit tests written in C++.

TODO:

- [x] It currently segfaults; debug
- [x] Make sure I haven't broken regular device adapter loading
- [x] Make sure adding this doesn't break or contaminate the SWIG bindings (MMCoreJ, pymmcore)

### DIFF
--- a/MMCore/LoadableModules/LoadedDeviceAdapter.cpp
+++ b/MMCore/LoadableModules/LoadedDeviceAdapter.cpp
@@ -29,40 +29,13 @@
 #include <memory>
 
 
-LoadedDeviceAdapter::LoadedDeviceAdapter(const std::string& name, const std::string& filename) :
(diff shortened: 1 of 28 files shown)


tlambert03:
oooo.  very exciting :)

PR #659, opened 2025-06-02
feat: add snap and sequence events

tlambert03:
This implements the following callbacks for image acquisition events, which are important for application-side reactions (particularly, knowing when a sequence acquisition has stopped)

```cpp
   virtual void onImageSnapped(const char* cameraLabel)
   {
      std::cout << "onImageSnapped() " << cameraLabel << '\n';
   }

   // numImages will be -1 if the sequence is infinite (i.e. startContinuousSequenceAcquisition)
   virtual void onSequenceAcquisitionStarted(const char* cameraLabel, long numImages, double intervalMs)
   {
      std::cout << "onSequenceAcquisitionStarted() " << cameraLabel << " numImages: " << numImages << " intervalMs: " << intervalMs << '\n';
   }

   virtual void onSequenceAcquisitionStopped(const char* cameraLabel)
   {
      std::cout << "onSequenceAcquisitionStopped() " << cameraLabel << '\n';
   }
```


marktsuchida:
Thanks, @tlambert03. I think there are a few things to think about here:

- `onImageSnapped`: should this be issued before (as currently written) or after starting to close the shutter? I think after, because we don't want to add arbitrary delays while the shutter is open. Note that it can be between the `SetOpen(false)` and the `waitForDevice()`.

- `onSequenceAcquisitionStarted`: would it be more symmetric with `onSequenceAcquisitionStopped` if this is also called from `CoreCallback::PrepareForAcq()` (so that both are called from a background thread)? In which case the name may be `onSequenceAcquisitionWillStart` (or `AboutToStart`), and it should probably be called before opening the shutter.

- `onSequenceAcquisitionStopped`: again, should this be called after starting to close the shutter?

- Should there also be an `onImageAboutToBeSnapped`, if only for symmetry?

Regarding the order relative to the autoshutter, it arguably depends on the use case. If these callbacks are used to change the state of the UI (e.g., a viewfinder display), it probably makes sense the way I propose (outside of the shutter control). But if the callbacks are used to do hardware control (as in an application-layer autoshutter), then the best ordering relative to the built-in autoshutter control is specific to the use case. However, in such cases the application could disable autoshutter  and do its own coordination in the callbacks.

tlambert03:
i moved both `onImageSnapped` and `onSequenceAcquisitionStopped` to after the shutter closed.  good points.

>  would it be more symmetric with onSequenceAcquisitionStopped if this is also called from CoreCallback::PrepareForAcq() (so that both are called from a background thread)? In which case the name may be onSequenceAcquisitionWillStart (or AboutToStart)

as for the position and naming of `onSequenceAcquisitionStarted`, i have a little less conviction on those.   I put them in the main core code partially because it wasn't clear to me whether every device adapter is guaranteed to call `GetCoreCallback()->PrepareForAcq`, and also because it seemed fine/natural to place that callback at the very end of a successful "spin up" of acquisition... after everything has successfully passed and returned.  I didn't use a "will start" style signal because i couldn't think of any action a callback would need to take "in preparation" for it to start (unlike something that needs to, say, do something before a device disappears).

same thing goes for `onImageAboutToBeSnapped` ... feels a little like YAGNI for now. so the symmetry feels nice, but perhaps hollow?

however, in both cases, I'm more than happy to just do those things if you would prefer it

marktsuchida:
> it wasn't clear to me whether every device adapter is guaranteed to call `GetCoreCallback()->PrepareForAcq`

It would be a device adapter bug if not -- autoshutter would not open otherwise. Plus, it is no more likely to be omitted than `AcqFinished`.

How would you use `onSequenceAcquisitionStarted` in the near term? Would this be where you start polling for new images to show in the preview?

In that case I think it might actually be a good thing to start the app-side image extraction loop before any frame is acquired, because at high frame rates the sequence buffer could more easily overflow if the loop takes its time starting up (e.g., because the caches are cold, the code hasn't been JITed, or whatever).

It will be good to think through what happens if the starting of the sequence acquisition somehow fails, however. If we issue the `Started` from `PrepareForAcq`, we need to guarantee that we somehow issue `Stopped` even if the acquisition in the camera never (successfully) starts. I don't remember of the top of my head what happens to autoshutter in such situations, but it needs to be dealt with somehow (and, at least in theory, we should be able to).

But if we issue `Started` from `StartSequenceAcquisition`, it is asynchornous with the camera, so there is no ordering guarantee between the `Started` and `Stopped` -- the latter could arrive before the former if the camera finishes the acquisition very quickly (successfully or otherwise).

I see your point about not needing `onImageAboutToBeSnapped`.

tlambert03:
updated and moved to `GetCoreCallback()->PrepareForAcq` and changed the signature to just include the device label.

> How would you use onSequenceAcquisitionStarted in the near term? Would this be where you start polling for new images to show in the preview?

yep

> It will be good to think through what happens if the starting of the sequence acquisition somehow fails, however. If we issue the Started from PrepareForAcq, we need to guarantee that we somehow issue Stopped even if the acquisition in the camera never (successfully) starts. I don't remember of the top of my head what happens to autoshutter in such situations, but it needs to be dealt with somehow (and, at least in theory, we should be able to).

do we need to verify this somehow before this should be merged?

marktsuchida:
> > It will be good to think through what happens if the starting of the sequence acquisition somehow fails, however. If we issue the Started from PrepareForAcq, we need to guarantee that we somehow issue Stopped even if the acquisition in the camera never (successfully) starts. I don't remember of the top of my head what happens to autoshutter in such situations, but it needs to be dealt with somehow (and, at least in theory, we should be able to).
> 
> do we need to verify this somehow before this should be merged?

After thinking about this a bit more, I think that the currently "correct" way for this to work is that cameras should call `AcqFinished` if they ever fail to start an acquisition after calling `PrepareForAcq`. In fact that is probably the only way, because `PrepareForAcq` can reasonably be called on the acquisition thread (so the Core cannot protect against such a failure based on the return value of `StartSequenceAcquisition`). So let's not worry about trying to deal with it here -- if it ever comes up, it will need to be fixed for that camera adapter.

PR #660, opened 2025-06-04
Fix OpenUC2 build for case sensitive file systems

jeskesen:
@beniroquai @marktsuchida 
I was getting build errors that the compiler couldn't find `openUC2.h`.  I noticed that it didn't exist, but `openuc2.h` did (same for .cpp).  I renamed the files and corrected the Makefile accordingly.

### DIFF
--- a/DeviceAdapters/OpenUC2/Shutter.h
+++ b/DeviceAdapters/OpenUC2/Shutter.h
@@ -3,7 +3,7 @@
 
 
 #include "DeviceBase.h"
-#include "openUC2.h"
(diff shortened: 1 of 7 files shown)


marktsuchida:
Thanks, @jeskesen! However, renaming files to change only the case tends to completely break git clones on Windows and Mac, in my experience. Could we keep the filenames and instead change the `#include` directives to match `openuc2.h`?

marktsuchida:
Thanks. Force-pushed to squash the reverted commits.

PR #672, opened 2025-06-27
fix: fix inconsistency on core XYStage Role, make setting roles more efficient

### LINKED ISSUE #671: Possible to have inconsistent state between core "current" and core "properties"
It appears possible to create an inconsistency between the Core properties and core roles "state".  Using `setProperty('Core',...` keeps them consistent, using `set<>Device()` does not

starting from the DemoConfig, if you call

```python
core.setXYStageDevice("")

print(core.getProperty("Core", "XYStage"))  # 'XY'  👀
print(core.getXYStageDevice())  # ''
```

however: 

```python
core.setProperty("Core", "XYStage", "")

print(core.getProperty("Core", "XYStage"))  # ''
print(core.getXYStageDevice())  # ''
```

tlambert03:
this fixes #671 

also tackles the `// TODO: more efficient` in all the `set<>Device` methods

### DIFF
--- a/MMCore/MMCore.cpp
+++ b/MMCore/MMCore.cpp
@@ -3419,8 +3419,8 @@ void CMMCore::setAutoFocusDevice(const char* autofocusLabel) throw (CMMError)
       currentAutofocusDevice_.reset();
       LOG_INFO(coreLogger_) << "Default autofocus unset";
    }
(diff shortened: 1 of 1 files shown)


marktsuchida:
Thanks!

PR #673, opened 2025-06-27
Add property type handling for CoreProperty 

tlambert03:
returns proper `getPropertyType` for core properties  (small `TODO`)


marktsuchida:
Good catch, thanks.

Issue #679, opened 2025-07-02
TODO:  Figure out how to save only user-defined labels for state devices to Config file


tlambert03:
In https://github.com/pymmcore-plus/pymmcore-plus/issues/478, it came up that the TriggerScopeMMTTL device creates a property "state", but unlike most state devices doesn't create a property 'Label':  

https://github.com/micro-manager/mmCoreAndDevices/blob/1352b67e2467b53cd0a6e4f1aa5adc9e4126ccdd/DeviceAdapters/TriggerScopeMM/TriggerScopeMMTTL.cpp#L128-L157

@nicost mentioned in https://github.com/pymmcore-plus/pymmcore-plus/issues/478#issuecomment-3028904514

> I believe I omitted these to avoid the lengthy list of labels in the configuration file for things that will most likely never be named. This device has 64000 states, so that will be a long list.

That was, in essence, taking advantage of the fact that if `GetPositionLabel` isn't defined, the label won't be saved.

https://github.com/micro-manager/mmCoreAndDevices/blob/1352b67e2467b53cd0a6e4f1aa5adc9e4126ccdd/MMCore/MMCore.cpp#L7228-L7237

It would be nice to have a better way to avoid verbose label storage.  It seems like what we really want is to avoid saving unnecessary labels: in other words, default labels that the user didn't create themselves (and which would be created by the device adapter code even without it being written in the config file).

Unfortunately, the "default" labels are created on a per-device ad hoc basis, using `SetPositionLabel` somewhere in initialize, like this:

https://github.com/micro-manager/mmCoreAndDevices/blob/1352b67e2467b53cd0a6e4f1aa5adc9e4126ccdd/DeviceAdapters/ASIFW1000/ASIFW1000.cpp#L386-L392

If we had a way to check the default label, we could avoid creating verbose config files without having to rely on non-standard device adapter implementations

PR #680, opened 2025-07-02
feat: Only save non-default (user-defined) labels in config

tlambert03:
One possible solution to #679 

just a thought, curious what @nicost and @marktsuchida think (not married to it of course... just an idea)

### DIFF
--- a/MMCore/MMCore.cpp
+++ b/MMCore/MMCore.cpp
@@ -779,6 +779,14 @@ void CMMCore::unloadDevice(const char* label///< the name of the device to unloa
    try {
       mm::DeviceModuleLockGuard guard(pDevice);
       LOG_DEBUG(coreLogger_) << "Will unload device " << label;
+      
+      // Clean up initial state labels for this device if it exists
+      auto it = initialStateLabels_.find(label);
+      if (it != initialStateLabels_.end()) {
+         initialStateLabels_.erase(it);
+         LOG_DEBUG(coreLogger_) << "Cleaned up initial state labels for device " << label;
+      }
+      
       deviceManager_->UnloadDevice(pDevice);
       LOG_DEBUG(coreLogger_) << "Did unload device " << label;
       
@@ -814,6 +822,10 @@ void CMMCore::unloadAllDevices() throw (CMMError)
       }
 
       LOG_DEBUG(coreLogger_) << "Will unload all devices";
+      
+      // Clear all initial state labels since all devices are being unloaded
+      initialStateLabels_.clear();
+      
       deviceManager_->UnloadAllDevices();
       LOG_INFO(coreLogger_) << "Did unload all devices";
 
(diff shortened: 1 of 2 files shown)


tlambert03:
For example...

Before this PR (demo configuration):

```ini
# Labels
Label,Dichroic,0,400DCLP
Label,Dichroic,1,Q505LP
Label,Dichroic,2,Q585LP
Label,Dichroic,3,89402bs
Label,Dichroic,4,State-4
Label,Dichroic,5,State-5
Label,Dichroic,6,State-6
Label,Dichroic,7,State-7
Label,Dichroic,8,State-8
Label,Dichroic,9,State-9
Label,Emission,0,Chroma-HQ620
Label,Emission,1,Chroma-D460
Label,Emission,2,Chroma-HQ535
Label,Emission,3,Chroma-HQ700
Label,Emission,4,89402m
Label,Emission,5,State-5
Label,Emission,6,State-6
Label,Emission,7,State-7
Label,Emission,8,State-8
Label,Emission,9,State-9
Label,Excitation,0,Chroma-D360
Label,Excitation,1,Chroma-HQ480
Label,Excitation,2,Chroma-HQ570
Label,Excitation,3,Chroma-HQ620
Label,Excitation,4,State-4
Label,Excitation,5,State-5
Label,Excitation,6,State-6
Label,Excitation,7,State-7
Label,Excitation,8,State-8
Label,Excitation,9,Empty
Label,Objective,0,Nikon 40X Plan Fluor ELWD
Label,Objective,1,Nikon 10X S Fluor
Label,Objective,2,Objective-2
Label,Objective,3,Nikon 20X Plan Fluor ELWD
Label,Objective,4,Objective-4
Label,Objective,5,Objective-5
Label,Path,0,State-0
Label,Path,1,State-1
Label,Path,2,State-2
Label,LED,0,Closed
Label,LED,1,385nm
Label,LED,2,470nm
Label,LED,3,550nm
Label,LED,4,635nm
Label,LED,5,State-5
Label,LED,6,State-6
Label,LED,7,State-7
Label,LED,8,State-8
Label,LED,9,State-9
```

After this PR

```ini
# Labels
Label,Dichroic,0,400DCLP
Label,Dichroic,1,Q505LP
Label,Dichroic,2,Q585LP
Label,Dichroic,3,89402bs
Label,Emission,0,Chroma-HQ620
Label,Emission,1,Chroma-D460
Label,Emission,2,Chroma-HQ535
Label,Emission,3,Chroma-HQ700
Label,Emission,4,89402m
Label,Excitation,0,Chroma-D360
Label,Excitation,1,Chroma-HQ480
Label,Excitation,2,Chroma-HQ570
Label,Excitation,3,Chroma-HQ620
Label,Excitation,9,Empty
Label,Objective,0,Nikon 40X Plan Fluor ELWD
Label,Objective,1,Nikon 10X S Fluor
Label,Objective,2,Objective-2
Label,Objective,3,Nikon 20X Plan Fluor ELWD
Label,Objective,4,Objective-4
Label,Objective,5,Objective-5
Label,LED,0,Closed
Label,LED,1,385nm
Label,LED,2,470nm
Label,LED,3,550nm
Label,LED,4,635nm
```

tlambert03:
Ah... because that auto population happens after Initialize()?  Can you point me to an adapter like that?

marktsuchida:
> Ah... because that auto population happens after Initialize()? Can you point me to an adapter like that?

I think the auto-population happens during initialization; an example is `ZeissCAN29`, for example the objective turret. So logically there would be no problem -- if nothing changes over time.

But things do change over time. One possibility is that the device adapter or driver decides to report a slightly different string (like "63x 1.4NA" instead of "63x NA1.4"). That would break any config groups referencing the state device's `Label` property. Perhaps this should be considered a rare situation and in any case the bug is in the adapter/driver.

The auto-detected labels could also change due to the microscopist actually swapping (or reordering) the components. If the labels are fixed in the config file, then the configuration would still be functional, albeit mislabeled. The labels can be edited in the HCW (I don't remember of config groups containing `Label` get automatically updated in this case, but conceivably they could be).

But if the labels are not stored in the HCW, then the config would be broken on loading and at least those config groups woul d need to be recreated from scratch to fix it.

I guess one "safe" workaround would be to omit the labels from the config file only when they are not only equal to default but also numerous (say, >100 states) -- though a messy heuristic.

nicost:
This looks so promising.  I I understand correctly, the outstanding work is:
- Ensure that there are no implicit labels included in config groups. If a label is referenced in a config group, it would always be included in the labels section of a config.
- Fix MMStudio and pymmcore-plus implementations of config saving to include "implicit" labels that are used in config groups into the labels section.
Is that correct?

PR #686, opened 2025-07-07
MMCore: Unset device role when unloading

marktsuchida:
This fixes an issue introduced in #674.

Regression test (for camera only, for now) added.

Before this change, an exception would be thrown when trying to unload the current camera, etc., because the property no longer had the current device's label as an allowed value on refreshing.

It's a bit of a messy fix, but the best we can do without major refactoring, I think.

An alternative way to fix this might be to disallow unloading devices currently assigned to a role, which may have been a more logical design, but I think that would break too much existing code.


marktsuchida:
Okay, the problem is actually this: although the `getCameraDevice()` family of functions return "" after a device has been unloaded, it does so only because the `weak_ptr` in `currentCameraDevice_` shows that the device no longer exists. But this doesn't work in the case of `unloadDevice()` because the latter keeps a `shared_ptr` (i.e., strong ref) to the device until after `properties_->Refresh()` is called.

I thought I could simplify the fix by ensuring that we release said `shared_ptr`, but actually that's not enough because other functions that call `unloadDevice()` may also hold a `shared_ptr` to the device (e.g. `unloadLibrary`). So I think proactively removing the device roles is a better fix. I'm going to force-push only to update the commit message.

PR #689, opened 2025-07-13
fix misc docs issues

tlambert03:
in pymmcore-nano, I'm experimenting with extracting docs from MMCore.  This fixes a few small warnings that crop up when you run doxygen on MMCore, and adds a test to prevent regressions.

Not actually adding a Doxyfile here, so as to avoid the impression that we're actually using it internally


marktsuchida:
Ok, I think the kinds of errors fixed in this PR will be detected by the setup in #788 (once we run it in CI).

Issue #642, opened 2025-05-16
List of small issues and PRs that need device interface version to be bumped


marktsuchida:
Major changes should be merged when they are ready, but we often have little changes that are not urgent and need the DIV to be incremented. It's nice to merge these on the same or close day as others, so that we keep low the frequency of version changes.

Here is a place to keep track of these.

---

- #588 and #640
- #558 
- #295 

----

- #595
- #596
- #722
- #754

PR #697, opened 2025-07-17
On shutterOpenChanged callback

nicost:
This PR lets Device Adapters inform the Core that the state of the shutter they control has changed.  It should be called from code that monitors the state of the shutter.  It would also be possible to call from code opening the shutter, but then you should make sure the shutter is really in the state reported (many shutters need some time to open / close).  

This PR updated the Device interface version as well as the minor version of the Core.  I am fine waiting to merge this PR until there are other reasons to increase the Device Interface version (as there are still a number of 3dparty device adpa\\apters that we can not automatically update).  @tlambert03

### DIFF
--- a/DeviceAdapters/DemoCamera/DemoCamera.cpp
+++ b/DeviceAdapters/DemoCamera/DemoCamera.cpp
@@ -3914,7 +3914,7 @@ int DemoShutter::OnState(MM::PropertyBase* pProp, MM::ActionType eAct)
 
       // apply the value
       state_ = pos == 0 ? false : true;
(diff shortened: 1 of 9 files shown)


nicost:
Did you have a chance to look at the @tlambert03 ?  OK to merge?

tlambert03:
sorry for the delay!  Thanks for the ping.  I'm totally fine/happy with `virtual void onShutterOpenChanged(const char* name, bool open)`

marktsuchida:
Thanks. I'll try to merge this (hopefully today or tomorrow), probably together with one or two items from #642.

PR #717, opened 2025-08-15
Remove camera "clear on buffer overflow" behavior

### LINKED ISSUE #640: Remove manual clearing of sequence buffer under `!stopOnOverflow`
Because #588 guarantees that `DEVICE_BUFFER_OVERFLOW` will not be returned from `InsertImage()` when acquisition should not stop, that error should now be treated the same way as any other error, and use of `GetCoreCallback()->ClearImageBuffer(this)` should be removed.

This should be a safe transformation to do on existing cameras, because the code to be removed is code that won't be executed any more anyway. We should just clean up once and for all.

marktsuchida:
Closes #640.

- **Remove "live" acquisition clear-on-overflow**

The special handling of `DEVICE_BUFFER_OVERFLOW` that is no longer necessary because of #588 is removed from all cameras.

- **Remove ClearImageBuffer() from Core callback**

And `ClearImageBuffer()` (deprecated in #588) is completely removed.

### DIFF
--- a/DeviceAdapters/ABS/ABSCamera.cpp
+++ b/DeviceAdapters/ABS/ABSCamera.cpp
@@ -1277,14 +1277,6 @@ int CABSCamera::InsertImage()
 
   int ret = GetCoreCallback()->InsertImage(this, pI, w, h, b, md.Serialize().c_str() );
 
-  if (!stopOnOverflow_ && ret == DEVICE_BUFFER_OVERFLOW)
-  {  
-    // do not stop on overflow - just reset the buffer
-    GetCoreCallback()->ClearImageBuffer(this);
-    // don't process this same image again...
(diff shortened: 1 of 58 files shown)


marktsuchida:
Camera devices where additional changes were made:
- AndorSDK3: removed `ClearImageBuffer()` call in `CCallBackManager::CPCStopLiveAcquisition()` (which is never called)
- Hikrobot: removed `ClearImageBuffer()` call at end of acquisition loop (likely a bug because the application may still be popping the images)

All other changes to cameras is a mechanical removal of the retry of `InsertImage()` after `DEVICE_BUFFER_OVERFLOW`, which should be a safe change. Several are missing error handling (on the return value of `InsertImage()`) but no attempt was made to fix that, to keep the risk of introducing bugs low.

PR #729, opened 2025-08-19
Experimental Meson build for 4 device adapters

marktsuchida:
DemoCamera, NotificationTester, SequenceTester, Utilities.

This uses a new convention for device adapter binary filenames: `XYZ.mmdev` instead of `[lib]mmgr_dal_XYZ{.dll,.so.0,}`.
(We can teach MMCore to accept both conventions by default.)


marktsuchida:
TODO: Run in CI.

PR #738, opened 2025-08-26
Add PropertyShutter utility device

alandolt:
This pull request introduces a new `PropertyShutter` utility device that can control any device property, not just digital-to-analog outputs as in `DAShutter` with Arduinos and NIDAQs. It allows using sliders or other properties as shutters, for example a halogen lamp that has only a voltage slider and no explicit on/off property. This makes it possible to use such devices as shutters in multi-dimensional acquisitions. The PropertyShutter is highly configurable, and the following properties can be set:

- Target Device & Property: which property the PropertyShutter will change
- Open & Closed Values: values to apply when the shutter is opened/closed
- Shutter Delay: optional delay after opening or closing
- Invert Logic: invert open/closed values if needed
- State: useful for debugging, allows manual control through the property panel

The implementation was successfully tested in real-life on a Zeiss Axio Imager Z2 microscope using the ZeissCAN29 device adapter. In this setup, the adapter exposes the halogen lamp only as a voltage slider without native shutter support. With the PropertyShutter, a pseudo shutter device is created that toggles the lamp on and off by controlling its voltage. Auto shutter control can be used as well, to automatically enable / disable the lamp if the camera is on / off. Further tests had been performed using the MMDemo Config. 

This adapter was compiled on Windows using MSVC v143.

### DIFF
--- a/DeviceAdapters/Utilities/Makefile.am
+++ b/DeviceAdapters/Utilities/Makefile.am
@@ -14,6 +14,7 @@ libmmgr_dal_Utilities_la_SOURCES = \
         MultiDAStateDevice.cpp \
         MultiShutter.cpp \
         MultiStage.cpp \
+        PropertyShutter.cpp \
         SerialDTRShutter.cpp \
         SingleAxisStage.cpp \
         StateDeviceShutter.cpp \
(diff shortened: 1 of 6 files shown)


marktsuchida:
Thank you, this is quite useful and cleanly done!

PR #739, opened 2025-08-29
Support TLV56x8 in the Arduino device adapter

asincerity:
This is a modified Arduino adapter designed to connect with the TLV56X8 chip for four-channel analog output. Corresponding adjustments to the code in the Arduino board are also required. The wiring method is as follows:
arduino     TLV56X8
5V    		VDD
GND		GND
DIGITAL 3   	SDI
DIGITAL 4 	SCK
DIGITAL 5   	CS1
DIGITAL 6 	CS2
The link to the Arduino board code is as follows:  https://github.com/asincerity/arduino_TLV56X8


asincerity:
"Thank you for your response. I have submitted the updated code and firmware in the PR, removing any Chinese characters. I think your idea of letting the code determine the number of DAs automatically is very cool. Here's a cost-effective 8-channel DAC board I found on AliExpress."

Issue #714, opened 2025-08-15
Incorrect debayering of leftmost and topmost pixels (PVCAM/MicroPublisher 6)


marktsuchida:
Folks in our lab are getting images that have the leftmost and uppermost 3 pixels looking incorrect:

<img width="690" height="605" alt="Image" src="https://github.com/user-attachments/assets/8fdec8f4-e8eb-4d2e-8200-fc71ab652616" />

PVCAM running a MicroPublisher 6 camera. I believe the debayering is done in the device adapter; it uses its own (modified) copy of `MMDevice/Debayer.{h,cpp}` (but we should probably also check that the generic version is correct).

Cc: @tomhanak

marktsuchida:
@tomhanak Thank you so much for looking into this.

I tend to agree that if we can avoid maintaining our own debayering routines, that would be a good thing (and these are probably not written in the most performant way, either). Maybe I can see if there is a way that we can use OpenCV's (or possibly LibRaw's) implementation without complicating the build/distribution too much (static linking would be ideal so that it doesn't clash with other device adapters). I'm also happy to set up the build dependencies if you ever want to move to your own (PVCAM SDK) implementation.

Or I might see if modifying the `GetPixel()`s (2 overloads) is enough to change the boundary condition (to reflected or replicated). This seems like the minimal fix at this time.

PR #744, opened 2025-09-08
PVCAM: Fix colors on image border after debayering

tomhanak:
I have reviewed full implementation, analyzed all 4 algorithms and added many comments with documentation and TODOs pointing to existing problems.

The original code seems to have swapped red and blue channels for two (out of four) bayer masks. It was compensated (not sure if intentionally) in the adapter code, where the only consequence was a mismatch between "Color - Sensor CFA" and "Color - Algorithm CFA" properties for some ROI sizes and positions. This is now fixed and documented in the code.

The default Nearest Neighbor algorithm is rewritten from scatch with only one iteration through the image plus extra handling on the edges and corners. Although it's not paralelized, it should be significantly faster than the original code because there is no branching in the loops.

The bilinear algorithm is slightly modified to fix the issues on image border. But no optimizations were done thus the result is probably a bit slower due to the fixes.

Remaining two Smooth Hue algorithms are documented, formatted, but not fixed.

This should fix issue #714.


marktsuchida:
Thank you so much, @tomhanak!

> The original code seems to have swapped red and blue channels for two (out of four) bayer masks. It was compensated (not sure if intentionally) in the adapter code

I wonder if perhaps arising from the fact that the image format for sending to the Core is BGRx rather than RGB -- I was a bit confused when I was taking a look at the MMDevice version of the code last week.

PR #553, opened 2025-01-24
Add Support for Controlling Three Elveflow's Devices in Micro-Manager

Vroouum:
- Implementation of device communication protocols to manage multiple devices of Elveflow (Ob1Mk4, MuxWirev3, MuxDistrib).
- Compatibility adjustments to ensure smooth operation across supported devices.
- Thorough testing to verify synchronization and performance when working with three devices concurrently.

PR #770, opened 2025-10-30
WIP: Storage device API cleanup

marktsuchida:
This PR builds on top of #535 (MM::StorageDevice). I'm using a new PR to keep things organized, since I expect to make some pretty extensive changes.
(Merging this PR should mark #535 as merged, as long as #535 is not further modified.)
(You will see all the commits of #535 in this PR, too.)

See only the [changes relative to #535](https://github.com/micro-manager/mmCoreAndDevices/compare/go2scope:mmstorage-draft...marktsuchida:mmstorage-mark).

This is a **work in progress** to refine the API design for Storage devices. The goal is to get to something that we can merge with reasonable confidence that major breaking changes won't be needed for further evolution.

Tasks (includes items in my review comments to #535; may not be exhaustive):

**Big issues**
- [ ] The API is overfitted to Go2ScopeTiff, with features probably not relevant to other backends
    - acquire-zarr needs all configuration at time of stream creation; after that, we only append pixel data
        - `AddImage()` is not possible
        - `configureDatasetDimension()` won't work after creation
        - `configureDatasetCoordinate()` won't work after creation (if ever)
        - The distinction between dimension "name" and "meaning" is probably not general
        - Summary metadata has no place to go (unless used as acquire-zarr "custom metadata", but iffy)
        - Image metadata has no place to go (perhaps we do want to support per-plane metadata in other backends, but it's not 100% clear what the best API would be -- a single string may not always suffice)
        - Custom metadata (mutable key-value API) has no place to go (acquire-zarr "custom metadata" is a single JSON blob)
    - When we consider the properties of other formats (e.g. OME-TIFF) or I/O libraries we'll have similar mismatches
    - Conversely, there are additional, backend-dependent configurations that can go into creating a dataset (e.g., everything in `ZarrStreamSettings`) that won't be possible with the current API
        - Not being able to specify chunking from the app seems critically limiting
    - Given that we're not even trying to come up with an API that abstracts backend differences (outside of pixel data), maybe just pass in a JSON configuration to `createDataset()` and leave up to storage backend to interpret?
        - This may allow us to e.g. pass in ~~yaozarrs-generated JSON~~ JSON auto-transformed on the app side from yaozarrs or ome-zarr-models in the case of acquire-zarr
        - The JSON config would replace all of: dataset (summary) metadata, dimension names, coordinate names, custom metadata; but `createDataset()` would still take `shape` and `pixelFormat` (possibly redundantly)
- [ ] _(reading)_ API should explicitly model write-only datasets (acquire-zarr): e.g., app should be able to query if readable

**API design**
- [x] Let the Core assign dataset handles, rather than the devices
- [x] `GetProgress()` - remove for now? (Also check for other unimplemented functions.)
- [ ] _(reading)_ What is the lifetime of the pointer returned by `GetImage()`? (Implicitly until next call?) Rather copy into caller-provided buffer?
    - Reading into caller-provided buffer better matches OS APIs and gives caller max control over allocation and copy (and caching, if desired; we certainly don't want to see each storage backend implement ad hoc caching)
    - Counterargument: some I/O libraries (TensorStore) may support caching and return a shared (refcounted) array; requiring a caller-provided buffer adds a copy. But we don't have a mechanism to support application (Python/Java) side handles to buffers, and adding that here would be scope creep. We can consider such an optimization in the future if it becomes important.
- [ ] _(reading)_ `getDeviceNameToOpenDataset()`: replace with `isDatasetSupportedByStorage(storageLabel, path)`
- [ ] `StorageDataType` should not be specific to storage; let's define a generally useful `PixelFormat` type
- [ ] `Create()` says dataset name "may be modified ... to avoid overwriting". This is unhelpful, especially when left to each device -- user has no way to predict the resulting dataset name. We've had lots of bad experience with this in the MMStudio API (and GUI).
- [ ] `IsOpen()` is questionable: Core may be able to support this now without device's help, but what is the use case? Remove.

**API technicalities**
- [ ] Sizes should be `size_t`, not `int`
- [ ] Axis lengths should probably be 64-bit
- [ ] Clarify rules for null terminator for metadata buffers
- [ ] _(reading)_ We need a way for `GetImage()` to return an error code
- [x] _(reading)_ `ReleaseStringBuffer()` implementation should not default to `delete[]`
- [ ] Consider using a custom type for handles (`struct DatasetHandle { int h; };`)

**API naming**
- [ ] `appendNextToDataset()`: rename to `popNextImageAndAppendToDataset()` (or similar)

**MMCore behavior**
- [ ] Only support mono images (for now), and correctly report error on RGB (BGRa)
- [ ] Multi Camera won't work with `appendNextToDataset()`; can we prevent misuse?
- [ ] `AppendImage()`: Core should throw if read-only, before asking device; similar for all invalid calls on read/write-only datasets
- [ ] Ensure datasets are closed when the storage device is unloaded

**Misc**
- [ ] `MaxMetadataLength` is unused; remove
- [ ] Can we have automated/unit tests in stead of ad-hoc programs?

**Functionality beyond write/read**
- [ ] _(reading)_ Clarify what (should) happen when user tries to "load" a dataset that is already open
- [ ] _(reading)_ Clarify `Delete()` requirements: dataset must be open (or user must pass path)
- [ ] _(reading)_ `List()` API is hard to use correctly (caller must initialize buffers with empty string; if all caller-provided buffers are filled, can't tell how much more is needed) -- and not exposed in Core API
- [ ] _(reading)_ `Freeze()` should either be required for read-write devices or (temporarily) removed from the API

Known limitations whose fixes are out of scope for this PR (can be fixed later):

- No direct stream ("attached storage") from camera to storage [yet](https://github.com/micro-manager/mmCoreAndDevices/pull/535#issuecomment-3464078761)
- AcquireZarrStorage only supports a single dataset open at a time

These plans are not set in stone, so please feel free to propose alternatives or ask questions!

Cc: @go2scope, @nicost, @tlambert03.


tlambert03:
i definitely think a step-wise, build up from minimal API approach is the right way to go for something this large.  It's nice to have the giant "working principle" branch to work with, but it's simply too large to evaluate.  Having you distill the essence of the core from the scope creep would be a great way to move forward on this.  
I think all you need for acquire-zarr support is `createDataset` with some way to map parameters to the ArraySettings spec there (this is tricky to do in a backend-agnostic way of course... so I think your backend-JSON approach is the only way to go for now),  `appendImageToDataset` and `closeDataset`.  (with the other things you mentioned being nice to have too)
I do think that ome-tiff is pretty important for probably more half of users, and i honestly don't see that going away any time soon with the pace of ome-zarr adoption and progression.

nenada:
Sorry I was out of the loop for the last couple of months and missed the API discussion. I am catching up now. @marktsuchida thank you for effort of integrating the storage API. Let me know what should I do to help. I am also available for meetings.

A couple of thoughts:
We don't need to worry about G2STIFF; it was the first adapter we chose to develop the API with, and it was the new format that no one used, so we had freedom to do whatever we wanted and try a bunch of approaches with the metadata. I was trying to come up with an API that made sense from the application side *in general*, without worrying about the idiosyncrasies of the existing formats.

We can remove go2scope and G2STiff  entirely from micromanager devices. I will add a separate dll with the g2stiff when the API becomes stable.

I agree with Mark's plan for incremental API development. Regarding all other issues, I think there is a lot of material and I do think it requires a meeting to discuss and coordinate. I can definitely put some work towards moving things forward, but not sure where to begin.

aliddell:
> I think @tlambert03 got things right wrt the acquire-zarr api. We have changed it a bit since #553. @aliddell maybe you can weigh in here?

Talley's assessment about what the aqz API needs is spot on. Several things in the (configuration) API have changed since #535, but probably the most relevant change is that multiple output arrays are now supported within a single Zarr dataset, so the properties of the dataset that pertain specifically to arrays have been consolidated into a `ZarrArraySettings` struct. There's also a flag which allows users to overwrite (or not) anything that lives in the base dataset.

PR #771, opened 2025-10-30
MMDevice: Delete an unused macro

marktsuchida:
Has multiple issues and should not be in DeviceBase. Doesn't even work since 2011.


marktsuchida:
This seems to have preemptively fixed a cryptic compile error with recent versions of Boost (which presumably do something with the symbol `_t`).

PR #782, opened 2025-12-03
IDSPeak: Add support for FPGA(Region0) binning engine

Pranjal264:
Added a 'BinningEngine' property to switch between Sensor and FPGA (Region0) binning. This allows users to use high binning factors (e.g., 4x, 8x) supported by the FPGA even if the sensor only supports 2x.

### DIFF
--- a/DeviceAdapters/IDSPeak/IDSPeakCamera.cpp
+++ b/DeviceAdapters/IDSPeak/IDSPeakCamera.cpp
@@ -72,7 +72,7 @@ const char* g_keyword_Peak_PixelFormat = "IDS Pixel Format";
 * The constructor does not access the hardware at all. The verification whether
 * these features are actually supported happens during the initialization.
 * Features are simply skipped if they turn out not to be supported.
-* 
+*
 * @params idx - Unique Camera id
 */
 CIDSPeakCamera::CIDSPeakCamera(int idx) :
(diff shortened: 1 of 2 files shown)


marktsuchida:
Thanks, @Pranjal264!

@Lars-Kool Would you be interested in reviewing this before we merge?

The only comment I had is that the commented out function (`InitializeBinning`) should be deleted (take a look at "Files changed" with "Hide whitespace" turned on under the Gear ⚙ menu).

Lars-Kool:
Thanks @Pranjal264 for taking the time to work on the IDS device adapter.

I have made some small modifications. The version proposed here assumes both binning engines (Sensor and Region0) are available on any camera, tries to set the binning engine, and catches the exception (relatively gracefully) if something goes wrong. I've added a couple lines to probe which engines are available, and only add those. This limits the amount of unsuccessful property change calls.

Besides that, I made some changes that did not modify the working of the code (like placing the new OnBinningEngine property with the other properties).

@marktsuchida What would be the best way to add my changes to this branch? I've tried fetching Pranjal's remote, which GitHub does not allow me to. I've tried pushing to Pranjal's fork directly, to which I don't have access either.

Besides that, it's ok by me. I tested it (after my changes) on one of my cameras (which sadly only supports sensor binning), and everything works as expected.

Pranjal264:
Thanks @Lars-Kool l for the review and for polishing the logic @marktsuchida for the oversight.

PR #807, opened 2025-12-24
DeviceBase: Remove OnXYStagePropertyChanged callbacks in code setting the XY stage positions.

### LINKED ISSUE #806: Confusing callback from CXYStageBase
When calling the function CXYStageBase.SetPositionUm(x, y) a Core Callback "OnXYStagePositionChanged(x, y) is automatically generated.  This is confusing, since at the moment that callback is generated, the stage is not (yet) at the desired position.  As a result, the UI jumps the position of the XY stage to the destination, then starts tracking the real location of the stage (if it uses callbacks or if the UI is polling.  

I vaguely remember discussion about this behavior.  I am now of the opinion this callback should be removed.  It will be useful to add a function "boolean ReportsPosition()" or similar to indicate if the device adapter itself calls back to the core to reports its position, an if it does not, the UI will know that it should query the position of the stage.  
@tlambert03 will that work for you?

nicost:
This is undoing commit d1962c6f8b which I made myself in 2011.  
Getting this callback automatically, even when the stage is not anywhere near the desired position has become annoying and confusing.

Closes #806. Will open another PR for the addition of a function to report whether or not the stage uses callbacks.

### DIFF
--- a/MMDevice/DeviceBase.h
+++ b/MMDevice/DeviceBase.h
@@ -1945,7 +1945,7 @@ template <class U>
 class CXYStageBase : public CDeviceBase<MM::XYStage, U>
 {
 public:
-   CXYStageBase() : originXSteps_(0), originYSteps_(0)
+   CXYStageBase() : originXSteps_(0), originYSteps_(0), xPos_(0), yPos_(0)
    {
       // set-up directionality properties
       this->CreateProperty(MM::g_Keyword_Transpose_MirrorX, "0", MM::Integer, false);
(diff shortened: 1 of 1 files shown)


marktsuchida (approved):
I also agree that this makes sense. Especially now that MMStudio StageControl supports polling.

In `micro-manager`, the following use this event; it might be good to make sure that none of them will become completely unusable.

- MMCache
- StageControl
- XYNavigator
- PositionListDlg
- HCS PlatePanel
- SnapOnMove plugin

I'm not too worried, because we're only talking about when the stage is moved by software command.

PR #808, opened 2025-12-24
MMDevice: adds functions to XYStage and Stage to let them signal whether they use callbacks

nicost:
let them signal whether they use callbacks to update the UI about their position.

Second part of issue #806.  I believe this will be useful once implemented (for instance, the Stage control window, Snap-on-Move, and the HCS plugin could all make use of it), but can easily be convinced otherwise.

### DIFF
--- a/MMCoreJ_wrap/pom.xml
+++ b/MMCoreJ_wrap/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>org.micro-manager.mmcorej</groupId>
     <artifactId>MMCoreJ</artifactId>
-    <version>11.13.0</version>
+    <version>11.14.0</version>
 
     <name>MMCore Java API</name>
     <description>Java bindings for MMCore, the device abstraction layer of Micro-Manager, the microscope control and acquisition platform.</description>


MMDevice/MMDevice.h
```diff
@@ -603,6 +603,15 @@ namespace MM {
...
       virtual int GetLimits(double& lower, double& upper) = 0;
 
+      /**
+      * @brief Stages can use the OnStagePositionChanged callback to signal
+      * updates about their position.  Some adapters do so, others do not,
+      * in which case the UI code should use polling.  This function signals whether
+      * the device adapters uses callbacks, so that the UI knows it does not need
+      * to poll this device
+      */
+      virtual bool UsesOnStagePositionChanged() = 0;
```
tlambert03: just curious: how do you intend to determine from the UI whether a specific device returns True on this... does it imply that you intend to add another public method to MMCore?

nicost: Very good point!  Completely forgot about that part.  Added now.  Please feel free to suggest better function names.

PR #817, opened 2026-01-16
Feature: fluorescent beads mode in DemoCamera

eirinnm:
This PR introduces a new mode to the Democamera, to go alongside "artificial waves" (the default) and "noise". 

The **Fluorescent Beads** mode simulates a sample containing fluorescent beads on a slide. It responds to virtual stage movement (XY panning and Z focusing):

[Screencast from 16-01-26 14:45:21.webm](https://github.com/user-attachments/assets/5002e653-2284-41b9-b1af-ca8f9ba97d74)

- Beads are procedurally generated using a tile-based system; pan the XY stage infinitely in any direction.
- Beads remain in consistent world coordinates. Only visible beads are stored in memory.
- Defocusing: gaussian blur is applied as Z moves away from zero.

#### BeadDensity
- **Range**: 10 to 500
- **Default**: 100
- **Units**: beads per tile (512×512 μm² area)
- **Description**: Controls the number of beads generated in each tile. Higher values create denser fields of beads.

#### BeadSize
- **Range**: 1.0 to 10.0
- **Default**: 2.0
- **Units**: micrometers (σ of Gaussian PSF)
- **Description**: Base size of the bead's point spread function at best focus. Typical fluorescent microspheres range from 0.2-10 μm diameter; this parameter controls the σ of the Gaussian that represents the bead.

#### BeadBrightness
- **Range**: 0.125 to 8.0
- **Default**: 1.0
- **Units**: relative intensity multiplier
- **Description**: Global brightness multiplier for all beads. Values >1 make beads brighter, values <1 make them dimmer. Independent of bit depth.

#### BlurRate
- **Range**: 0.1 to 1.0
- **Default**: 0.5
- **Units**: σ (pixels) per μm of defocus
- **Description**: Rate at which beads blur when out of focus. The defocus blur is calculated as:
  ```
  σ_defocus = BlurRate × |Z_position|
  ```
  Higher values cause faster blurring as you move the Z-stage away from focus (Z=0).

----
### Known limitations
The XY panning and Z defocusing relies on the devices being called "XY" and "Z", which is the case for the default demo configuration.

### DIFF
--- a/DeviceAdapters/DemoCamera/DemoCamera.cpp
+++ b/DeviceAdapters/DemoCamera/DemoCamera.cpp
@@ -392,7 +392,7 @@ int CDemoCamera::Initialize()
    pAct = new CPropertyAction(this, &CDemoCamera::OnBeadBrightness);
    nRet = CreateFloatProperty("BeadBrightness", beadBrightness_, false, pAct);
    assert(nRet == DEVICE_OK);
-   SetPropertyLimits("BeadBrightness", 0.5, 2.0);
+   SetPropertyLimits("BeadBrightness", 0.125, 8.0);
    
    pAct = new CPropertyAction(this, &CDemoCamera::OnBlurRate);
    nRet = CreateFloatProperty("BlurRate", blurRate_, false, pAct);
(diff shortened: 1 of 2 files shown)


eirinnm:
Thanks! If you prefer, I can simplify it a bit - removing the user parameters like beadSize and just going with sane defaults.

marktsuchida:
My thing that @nicost mentioned is https://github.com/marktsuchida/mmdev-SimulatedCamera, which draws filaments and supports XYZ movement (and simulates shot+read noise). I do plan to move it to this org (micro-manager) and ship it with nightly builds once I figure out how to add a mechanism to package out-of-tree device adapters (integrating it into the existing build systems would be a lot of work due to dependencies). (See its `justfile` to see how to build.)

PRs there would be welcome, of course. Particles ("disks", since it's 2D rendering) would be a nice addition.

Or if you feel it better fits in DemoCamera, that's fine by me, too.

nicost:
Thanks @eirinnm !

PR #823, opened 2026-01-30
Address AlliedVisionCamera issues

ischoegl:
I am facing hard crashes when running an AlliedVision Prosilica GT2000-NIR camera, and have also observed some other driver malfunctions. I am interfacing with the camera via `pymmcore`.

- [x] `CMMCore.getROI()` always yields zeros -> fixed in `AlliedVisionCamera::GetROI`
- [x] `CMMCore.setExposure()` occasionally fails to set correct value (_Edit: traced back to round-off error in conversion from `float` to micro-seconds._)
- Deferred to separate PR: `CMMCore.snapImage()` results in occasional hard crashes (sometimes hundreds of captures in), specifically C++ segmentation fault (exit code 0xC0000005) in `pymmcore`

### DIFF
--- a/DeviceAdapters/AlliedVisionCamera/AlliedVisionCamera.cpp
+++ b/DeviceAdapters/AlliedVisionCamera/AlliedVisionCamera.cpp
@@ -434,8 +434,15 @@ double AlliedVisionCamera::GetExposure() const
 
 void AlliedVisionCamera::SetExposure(double exp_ms)
 {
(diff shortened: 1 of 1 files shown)


ischoegl:
@marktsuchida … thanks for the pointer! I have the dll compiled locally already (including an apparently failed attempt to harden against potential issues), and will most definitely try the Visual Studio debugger. I come from the Linux/macOS side but should be able to figure this out. Once I do, I will either post a fix or an issue report.

PR #829, opened 2026-02-09
iSIM Device Adapter

kmdouglass:
This is a pull request for a device adapter for constructing analog output waveforms from a NIDAQ board for an instant structured illumination microscope (iSIM). It originated with a thread between myself, @dpshepherd, and @nicost  here: <https://forum.image.sc/t/custom-waveform-generation-for-ni-daq-galvo-control/118662>

The main difference with the pre-exisiting NIDAQ device adapter is that it constructs and outputs analog waveforms for the iSIM's camera, galvanometer mirror, and AOTF. The NIDAQ device adapter, on the otherhand, acts as a state machine by outputting predetermined signals corresponding to each state. The galvo waveform for the iSIM in particular is somewhat complex; it is composed of linear ramps and constant parking periods that need tight synchronization with the camera's exposure.

The device adapter currently works for Live and Snap imaging by wrapping a physical camera to intercept some relevants calls into the core, much like the [TeensyPulseGenerator](https://github.com/micro-manager/mmCoreAndDevices/tree/main/DeviceAdapters/TeensyPulseGenerator). There are still some questions I have regarding hardware sequencing before it fully meets my requirements.

A few points:

- This device adapter is highly coupled to our particular setup. I think it could serve as a starting point for a more general device adapter. This would require quite a lot of discussion.
- I put an interface called `IDAQAdapter` between the device adapter code and the actual calls to the NIDAQmx library. This allows mocking out the hardware to test waveform construction logic.
- [As I explain here](https://kylemdouglass.com/posts/triggering-analog-voltage-outputs-using-a-ni-daq-counter/), the triggering logic is a bit indirect but necessary to avoid having to reload the waveforms in software after each frame. I use a counter channel on the NIDAQ that can be triggered by a TTL input, and this counter serves as the clock for the analog waveforms.

## Help Needed

I need help understanding how to make this compatible with hardware sequencing. Currently, I have only implemented what I call "normal" mode imaging. This means that if an AOTF MOD IN channel is enabled, then light from that channel will illuminate the sample on each frame.

What I would like is to do two color (or more) interleaved imaging. I know how to construct the waveforms for this, but I do not currently know how the information about a multichannel MultiD Acquisition is passed into the device. I would need this information at the device adapter level to construct the waveforms.

Below is an example of the desired waveforms for three color interleaved imaging.

image

I'd be happy to discuss further and to answer any questions you might have.

## To Dos

- [x] Implement hardware sequencing for multi-channel MDAs
- [x] ~Alert user when the exposure time setting is too short to allow for all rows to expose simultaneously due to the camera's rolling shutter~ I can't do this because `SetExposure` returns `void`, so I clamp the requested exposure time instead.
- [x] Read back exposure time from camera after setting it
- [x] ~Related to the above, decide what to do about the initial exposure time in the GUI (10 ms) being shorter than the readout time~ This is both a configuration issue and a problem with PVCAM, not iSIMWaveforms. Setting the starting exposure time for PVCAM in the configuration file so that it's longer than the readout time at startup solves it. I will need to document this later.
- [x] Discover PFI channels for the trigger port so that the user selects which one to use (instead of typing in the PFI channel)
- [x] Decouple the device adapter from PVCAM's readout time property
- [x] ~Figure out what to do about hard-coded NIDAQmx include and dependency paths in the VS solution file~ @marktsuchida Will take care of this when merging. Note that I'm currently testing against NIDAQmx 2026 Q1.
- [x] Add License.txt
- [x] Remove Makefile.am
- [x] Remove blanking output if no MOD IN channels are enabled and we're in Snap or Live mode
- [x] ~Rebuild waveforms whenever the ROI size changes because the readout time will change~ No longer needed since `SyncTimingFromCamera()` was implemented.

### DIFF
--- a/DeviceAdapters/iSIMWaveforms/iSIMWaveforms.vcxproj
+++ b/DeviceAdapters/iSIMWaveforms/iSIMWaveforms.vcxproj
@@ -57,12 +57,12 @@
       <Optimization>Disabled</Optimization>
       <PreprocessorDefinitions>_DEBUG;_WINDOWS;_USRDLL;MPBLASER_EXPORTS;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <SDLCheck>true</SDLCheck>
-      <AdditionalIncludeDirectories>C:\Program Files (x86)\National Instruments\Shared\ExternalCompilerSupport\C\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>$(MM_3RDPARTYPRIVATE)\NationalInstruments\DAQmx_9.2\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
-      <AdditionalLibraryDirectories>C:\Program Files (x86)\National Instruments\Shared\ExternalCompilerSupport\C\lib64\msvc;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalLibraryDirectories>$(MM_3RDPARTYPRIVATE)\NationalInstruments\DAQmx_9.2\lib64\msvc;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
       <AdditionalDependencies>nidaqmx.lib;%(AdditionalDependencies)</AdditionalDependencies>
     </Link>
   </ItemDefinitionGroup>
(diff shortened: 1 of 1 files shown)


nicost:
Looks very good!  Dying to hear if it works as expected!

> Just a point of clarification: I'm actually using analog outputs for the AOTF signals and not digital outputs because our synthesizer accepts a 0 - 10 V analog signal whose value determines the laser intensity (more specifically the diffraction efficiency). But I doubt this matters for making the hardware sequencing work.

Indeed.  If you want to be able to output different voltages (controlled by the UI), you would need to implement something different than a StateDevice, but if it is all just 0 or 10V, a StateDevice should be fine.

DeviceAdapters/iSIMWaveforms/iSIMWaveforms.cpp
```diff
@@ -0,0 +1,2295 @@
...
+const char* g_DeviceName = "iSIM Waveforms";
+const char* g_DeviceDescription = "Generates analog waveforms for an iSIM microscope";
+const char* g_Undefined = "Undefined";
+const char* g_PhysicalCamera = "Physical Camera";
+const char* g_IllumDeviceName = "iSIM Illumination Selector";
+const char* g_IllumDeviceDesc = "Selects which AOTF MOD IN channel is active for each frame";
+const char* g_ReadoutTimeNone = "None";
+
+// Set to true to use MockDAQAdapter instead of the real NIDAQmx adapter
+static constexpr bool kUseMockDAQ = false;
```
kmdouglass: This is only used for testing. I'm not sure whether you have compile-time flags for this sort of thing or not. It should always be false for release builds.

DeviceAdapters/iSIMWaveforms/iSIMWaveforms.cpp
```diff
@@ -0,0 +1,2295 @@
...
+	std::vector<std::string> devices = daq_->getDeviceNames();
+	CPropertyAction* pAct = new CPropertyAction(this, &iSIMWaveforms::OnDevice);
+	std::string defaultDevice = devices.empty() ? "" : devices[0];
+	deviceName_ = defaultDevice;
+	CreateStringProperty("Device", defaultDevice.c_str(), false, pAct, true);
+	for (const auto& dev : devices)
+		AddAllowedValue("Device", dev.c_str());
+
+	// Discover channels for default device
+	availableChannels_ = daq_->getAnalogOutputChannels(deviceName_);
```
kmdouglass: This is the only device communication that I do in the constructor. I discover all the available DAQ devices and their available analog output channels. The NIDAQ device adapter, for comparison, only discovers DAQ devices.

marktsuchida:
I pushed a commit that points the project to our internal copy of DAQmx 9.2 (old, but shouldn't matter for building what's used here -- it will use the locally installed version at run time). Should be ready to merge.

When working on the code, you can either edit the project properties temporarily, or you can put a copy (or symlink, if enabled) of the DAQmx include and lib64 directories where the build points. `$(MM_3RDPARTYPRIVATE)` points to `../../3rdparty` relative to `mmCoreAndDevices` root.

DeviceAdapters/iSIMWaveforms/iSIMWaveforms.cpp
kmdouglass: Fair point. I'll do this in a follow up as you suggest.

Issue #797, opened 2025-12-14
Move ImageMetadata.h from MMDevice (back) to MMCore


marktsuchida:
`MMDevice/ImageMetadata.h` defines the `Metadata` class and associated classes (`MetadataSingleTag`, `MetadataKeyError`, etc.).

`Metadata` is a somewhat confusing class because it's a key-value container, but the keys can be either (device label + tag) or just tag, except that half the time the device + tag is treated as a single string. And it _appears to_ contain support for "array tags", which were never fully implemented (probably a good thing because I have issues with its design, but I digress). There is also a "read only" flag, also never fully supported.

It currently performs double duty:

- It is the container for camera devices to pass per-frame metadata to the Core (in serialized form). It is also used internally in `CCameraBase` for the `AddTag()`/`RemoveTag()`/`GetTags()` mechanism used by Multi Camera et al.

- It is the container for CMMCore to provide per-frame metadata for sequence acquisition images. In MMCoreJ it is more standard to use `TaggedImage`, but `Metadata` and related classes are also exposed and used in the MMCoreJ API. They are also exposed in the pymmcore API.

This double role makes it hard to make changes and improvements, because any change will potentially affect both camera device source compatibility (not the device interface version, [any](https://github.com/micro-manager/mmCoreAndDevices/pull/575) [more](https://github.com/micro-manager/mmCoreAndDevices/pull/710)) and the MMCore API at the same time. Also, MMDevice and MMCore could use _different_ potential improvements. The fewer ways in which MMDevice constrains MMCore's API, the better, and the `Metadata` classes are the worst offender (the others are mostly just constants).

There are additional issues with the `Metadata` class being used in MMDevice:

- Some of the functions throw exceptions. This is not an appropriate design for something used in device adapters. We get away with it because use cases that could cause errors is rare.

- The license header says `ImageMetadata.h` is part of MMCore and is LGPL-licensed. This is purely an error we can correct, but betrays the origin of the code (explains why it's poorly matched to MMDevice's needs).

'Nuff said. So I propose that we move `ImageMetadata.h` back to MMCore, where it can continue to serve as an API type in MMCoreJ/pymmcore (and evolve or get deprecated as needed).

We would then need a replacement for MMDevice's purposes. In fact, I suggest we first add a replacement, then convert all cameras to use the replacement, then finally perform the moving of `ImageMetadata.h`.

The requirements for the replacement (let's call it `MM::MetadataMap`):

[...]

PR #830, opened 2026-02-09
Deprecate Metadata.put()

marktsuchida:
Formally deprecate `Metadata.put()`, which was already noted "deprecated", and convert all usage to the exactly equivalent `PutImageTag()`.

This does not change any behavior but simplifies working on #797 (more greppable, and one fewer pattern to worry about).

Keeping `put()` as deprecated instead of removing it outright, because removing it would cause a breaking change to the MMCore API (even though use is likely rare).


marktsuchida:
> Keeping `put()` as deprecated instead of removing it outright, because removing it would cause a breaking change to the MMCore API

Well, sort of. Neither `put()` not `PutImageTag()` is exposed via MMCoreJ and pymmcore because they are member function templates and we don't instantiate them in the SWIG wrappers. Maybe we can remove.

PR #850, opened 2026-02-19
XY stage position callbacks

ieivanov:
Follow up to #807

In this PR, I'm updating the `OnXYStagePositionChanged` callback on the `CDemoXYStage` to report the **final** position rather than the starting position. 

I'm also adding a `OnXYStagePositionChanged` callback for the ASITiger XY stage, which currently does not emit this callback. I haven't tested this with hardware, but I think it looks like a pretty straightforward change which does not modify the original behavior of relying on `CXYStageBase<CXYStage>::SetPositionUm(x, y)`.

### DIFF
--- a/DeviceAdapters/iSIMWaveforms/MockDAQAdapter.h
+++ b/DeviceAdapters/iSIMWaveforms/MockDAQAdapter.h
@@ -109,10 +109,6 @@ public:
             logger_("[MockDAQ] clearTasks()");
     }
 
-private:
-    LogCallback logger_;
-    size_t channelCount_ = 0;
-
     std::vector<std::string> getDeviceNames() const override
     {
         return {"MockDev1", "MockDev2"};
@@ -150,4 +146,8 @@ private:
             return {"/MockDev2/PFI0", "/MockDev2/PFI1"};
         return {};
     }
+
+private:
+    LogCallback logger_;
+    size_t channelCount_ = 0;
 };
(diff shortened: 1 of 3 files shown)


marktsuchida:
Perhaps a slight misunderstanding here — the aim of #807 was to avoid notifying the final position at the start of the move (note that the move does not complete during the SetPosition call), because that causes GUI displays to jump to the destination and back to an intermediate point, if the stage already notifies during the move — which I think the Demo stage does, or at least that's what the code looks like.

On the other hand, it's true that ASITiger does not currently notify position changes at all (since #807, like many stages). @nicost Do you think it makes sense for stages to notify their destination at the start of a move, when asynchronous notifications are not immediately available?

Also Cc @jondaniels, @bls337.

ieivanov:
Ah, sorry, maybe I've misunderstood the purpose of this callback and how to use it. 

In `DeviceBase.h` we have 

```c++
   /**
    * @brief Signal that the XY stage has arrived at a new position.
    */
   int OnXYStagePositionChanged(double xPos, double yPos)
   {
      if (callback_)
         return callback_->OnXYStagePositionChanged(this, xPos, yPos);
      return DEVICE_NO_CALLBACK_REGISTERED;
   }
```

and then again in `MMDevice.h`

```c++
      /**
       * @brief Signal the UI when the XY stage has reached a new position.
       */
      virtual int OnXYStagePositionChanged(const Device* caller, double xPos, double yPos) = 0;
```

Based on this, it seems to me that `OnXYStagePositionChanged` should emit once the stage has reached its final destination and report on these coordinates. In `CDemoXYStage::SetPositionSteps` this call comes after `timeOutTimer_ = new MM::TimeoutMs(currentTime, moveDuration_ms_);` which made be think that the move is now complete, but I think that's not the case.

Is there a different callback that emits once the stage has reached its final destination after a move?

ieivanov:
Fantastic, thanks for clarifying and proving context here. I'll close this PR.

PR #860, opened 2026-02-20
MMDevice: remove 3 Camera functions (bumps DIV)

### LINKED ISSUE #595: `PrepareSequenceAcqusition` should be removed
It looks like a number of cameras implement `PrepareSequenceAcqusition` (sic) by calling `GetCoreCallback()->PrepareForAcq(this)`, which opens the shutter when autoshutter is enabled. Given that `CMMCore::prepareSequenceAcquisition` simply calls the device's `PrepareSequenceAcqusition` (and is not guaranteed to be followed by `startSequenceAcquisition` (and even when it is, the latter may fail before scheduling a shutter close)), this is incorrect.

The intended behavior of `PrepareSequenceAcqusition` (according to the comment on `CMMCore::prepareSequenceAcquisition`) was to commit the current settings to the camera, or otherwise make preparations, so that a subsequent call to `StartSequenceAcquisition` will be faster. It looks like very few, if any, camera adapters actually do this (correctly, at least).

Many other cameras, including the `CCameraBase` default, just return `DEVICE_OK`. This is harmless.

It also turns out that mmstudio and acqEngine (Clojure) currently do not call `prepareSequenceAcquisition` anywhere. AcqEngJ and pymmcore-plus each do in one place, but they probably don't rely on the call doing anything.

We already have a mechanism to avoid slow preparation after opening the autoshutter (such preparation should be done in `StartSequenceAcquisition` but before calling `PrepareForAcq`), so any advantage offered by (a correctly implemented) `PrepareSequenceAcqusition` would only be gained if the call were made well in advance, or in parallel with other motion, etc.

Also, among such preparations, reallocating the sequence buffer (V1) can be done explicitly by the application (`initializeCircularBuffer`). So `PrepareSequenceAcqusition` should have really been only about committing settings to the camera hardware. Again, this is rarely if ever done. And #593 proposes a new way to handle this (separate Arm + Start).

Given all this, I think we should just deprecate this function in `CMMCore`, make it a no-op, and remove it from MMDevice.

(Implementations in camera adapters can then be deleted in most cases if not called from within the adapter itself.)

Related: #453.

### LINKED ISSUE #596: `MM::Camera::GetPixelSizeUm()` should be removed from MMDevice
This function is useless and never called from MMCore.

Noted by @henrypinkard in #590.

### LINKED ISSUE #754: MM::Camera::GetComponentName should be removed
This is another case similar to #596 -- a function in the Camera device interface that is never called by MMCore, where camera adapters that do override it just do whatever.

marktsuchida:
Closes #596
Closes #754
Closes #595

These have already been made `final` recently to prevent override.
`GetPixelSizeUm()` - never called from Core, intent unclear
`GetComponentName()` - never called from Core, inconsistent impls
`PrepareSequenceAcqusition()` - inconsistent impls, seldom of value, typo in name

### DIFF
--- a/MMDevice/DeviceBase.h
+++ b/MMDevice/DeviceBase.h
@@ -1422,21 +1422,9 @@ public:
     */
    virtual int StopSequenceAcquisition() = 0;
 
-   // It appears that this function was never used by MMCore and is slated for
-   // removal. Concrete cameras should not override.
-   double GetPixelSizeUm() const final { return 0.0; }
-
    virtual unsigned GetNumberOfComponents() const
    {
       return 1; // Default to monochrome (ie not RGB)
-   }
-
-   // To be removed (never used by MMCore); devices should not override.
-   virtual int GetComponentName(unsigned channel, char* name) final
-   {
-      (void)channel;
-      CDeviceUtils::CopyLimitedString(name, "");
-      return DEVICE_OK;
    }
 
    /**
@@ -1493,9 +1481,6 @@ public:
       }
       CDeviceUtils::CopyLimitedString(serializedMetadata, md.Serialize());
    }
-
-   // To be removed; devices should no longer override.
-   virtual int PrepareSequenceAcqusition() final {return DEVICE_OK;}
 
    /**
     * @brief Start sequence acquisition.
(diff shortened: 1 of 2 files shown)


marktsuchida:
This will be merged as part of #861.

PR #864, opened 2026-02-24
Update ScientificaMotion8.cpp

NigelWagg:
Change of return value in below methods to return is_moving instead of ret bool M8XYStage::Busy()
bool M8ZStage::Busy()


NigelWagg:
Change of return value in below methods to return is_moving instead of ret bool M8XYStage::Busy()
bool M8ZStage::Busy()

PR #865, opened 2026-02-24
Change IsMoving return value to is_moving

NigelWagg:
Updated return value in Busy method to return is_moving instead of ret.


NigelWagg (commented):
Change to Busy to return IsMoving value

Issue #498, opened 2024-10-09
We should have the option to forward Core logging to an app-provided callback


marktsuchida:
MMCore is a library, and libraries that do their own logging are generally problematic. When being used from an environment (Java, Python, other app framework) that has its own logging, the best architecture would be that MMCore plugs into the app's logging strategy.

Something like the ability to specify a callback as the (sole) destination for log messages. (It would also need to be synchronous; any log buffering should be up to the app.)

See: https://github.com/pymmcore-plus/pymmcore-plus/issues/385

Probably not a high priority immediately but a possible feature.

Possible downside: if the app is in Python, this may mean acquiring the GIL on every log entry from devices. May have performance implications.

PR #906, opened 2026-03-25
MMCore logging: add log rotation, full log level/source support

marktsuchida:
The goal here is to make MMCore's logging usable as a destination for logging from Python or SLF4J. The main motivation is for pymmcore-plus (https://github.com/pymmcore-plus/pymmcore-plus/issues/385).

This is the opposite direction to what I proposed in #498. But given that the MMCore logging is already non-blocking and (relatively) high-performance, it makes more sense to send logging to MMCore in order to get unified logging. (Sending MMCore logging to Python logging would require processing every log entry with the Python GIL held.)

APIs added here:

- `mmcore::LogLevel` enum (`Trace, Debug, Info, Warning, Error, Critical`, which is the union of SLF4J and Python logging)
- `setPrimaryLogFileRotation(long long maxFileSize, int maxBackupCount)` (disabled by default for compatibility)
- `log(const char* msg, mmcore::LogLevel level)`
- `log(const char* msg, mmcore::LogLevel level, const char* loggerName)`
- `setPrimaryLogLevel(mmcore::LogLevel level)`
- `getPrimaryLogLevel() -> mmcore::LogLevel`

Probably should be tested in Java/Python a bit before merging.

### DIFF
--- a/MMCore/Devices/DeviceInstance.cpp
+++ b/MMCore/Devices/DeviceInstance.cpp
@@ -35,8 +35,8 @@ namespace mmcore {
 namespace internal {
 
 using mmcore::internal::logging::Logger;
-using mmcore::internal::logging::LogLevelDebug;
-using mmcore::internal::logging::LogLevelInfo;
+using mmcore::LogLevelDebug;
+using mmcore::LogLevelInfo;
 
 int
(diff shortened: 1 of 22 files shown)


marktsuchida:
Now also colors the level (severity) code when sending the log to stderr.

PR #907, opened 2026-03-26
Enable digital-output shutter control without camera synchronization using the NIDAQ adapter + Fix Labels in NDIAQ adapter

alandolt:
There are situations in which a triggerable light source must be used without a camera synchronization signal, for example, when the camera does not provide a sync output, in legacy setups, or when synchronization becomes impractical due to additional hardware such as a DMD. In these cases, hardware blanking must be disabled and the light source controlled directly via a digital output (you may ask why in this case go over hardware TTL control instead of software on / off control => hardware control is much faster and much more reliable, especially for old Lumencor Spectra's)

Previously, the NIDAQ adapter did not function correctly when hardware blanking was set to off. Opening or closing the shutter failed to drive the digital output back to 0 because the internal logic implicitly relied on blanking. With these two commits, hardware blanking can now be disabled explicitly, and the configured digital output port is used to directly enable or disable the light source. When live mode is exited, the output is reliably reset to 0, ensuring that the light source is switched off.

To support this behavior, the DigitalOutputPort class was extended with explicit gate (shutter) control. New SetGateOpen and GetGateOpen methods manage the internal gate state and apply it to the hardware as appropriate. The OnState logic was refined to distinguish between hardware-timed (blanking) and software-timed operation, ensuring that gate state changes are only applied in software-timed mode and do not interfere with hardware blanking.

During testing, this workflow behaved correctly with an Andor Zyla camera, which properly manages shutter control, but not with Teledyne PVCAM-based cameras. To address this discrepancy, the PVCAM adapter was updated to allow explicitly switching off the light source. In addition, abortAcquisitionInternal in PVCAMUniversal.cpp now notifies the core system when an acquisition finishes, allowing the AutoShutter to close the shutter automatically and restoring consistent shutdown behavior across camera backends.

Edit: Latest commit adds labels of NIDAQ devices to property browser (you can set-up labels during the hardware setup, but they weren't shown in the property browser during runtime) => fixed.

### DIFF
--- a/DeviceAdapters/NIDAQ/NIDAQ.h
+++ b/DeviceAdapters/NIDAQ/NIDAQ.h
@@ -470,7 +470,6 @@ public:
 
     // Gate (shutter) interface
     int SetGateOpen(bool open);
-    int GetGateOpen(bool& open);
 
     // action interface
     // ----------------
@@ -499,7 +498,6 @@ private:
     bool sequenceRunning_;
     bool blanking_;
     bool blankOnLow_;
-    bool open_;
     long pos_;
     long numPos_;
     long highestLabeledPos_;
--- a/DeviceAdapters/NIDAQ/NIDAQ.h
+++ b/DeviceAdapters/NIDAQ/NIDAQ.h
@@ -422,7 +422,51 @@ public:
     virtual void GetName(char* name) const;
     virtual bool Busy() { return false; }
 
-    virtual unsigned long GetNumberOfPositions()const { return numPos_; } // replace with numPos_ once API allows not creating Labels
+    // For 8-bit ports, all states are enumerable. For wider ports, only
+    // enumerate positions that have been explicitly labeled (e.g. via config
+    // file). This prevents MMCore from iterating billions of states.
+    // The State property limits still allow the full value range.
+    static const long maxEnumerablePositions_ = 65536;
+    virtual unsigned long GetNumberOfPositions()const {
+       if (numPos_ <= 255)
+          return numPos_ + 1;
+       return (highestLabeledPos_ >= 0 && highestLabeledPos_ < maxEnumerablePositions_)
+          ? (unsigned long)(highestLabeledPos_ + 1) : 0;
+    }
+
+    // Return the numeric string as default when no explicit label exists.
+    // This prevents errors when the Label property is read for unlabeled states.
+    virtual int GetPositionLabel(long pos, char* label) const
+    {
+       int ret = CStateDeviceBase<DigitalOutputPort>::GetPositionLabel(pos, label);
+       if (ret != DEVICE_OK)
+       {
+          CDeviceUtils::CopyLimitedString(label, std::to_string(pos).c_str());
+          return DEVICE_OK;
+       }
+       return ret;
+    }
+
+    // Override to fill in default labels for gaps when labels are added
+    // beyond the auto-generated range (e.g. from config file on wide ports).
+    virtual int SetPositionLabel(long pos, const char* label)
+    {
+       // Fill in default labels for any positions below pos that have no label
+       if (pos > highestLabeledPos_ && pos < maxEnumerablePositions_)
+       {
+          char buf[MM::MaxStrLength];
+          for (long i = highestLabeledPos_ + 1; i < pos; i++)
+          {
+             // Only fill if no label exists yet
+             if (CStateDeviceBase<DigitalOutputPort>::GetPositionLabel(i, buf) != DEVICE_OK)
+             {
+                CStateDeviceBase<DigitalOutputPort>::SetPositionLabel(i, std::to_string(i).c_str());
+             }
+          }
+          highestLabeledPos_ = pos;
+       }
+       return CStateDeviceBase<DigitalOutputPort>::SetPositionLabel(pos, label);
+    }
 
     // Gate (shutter) interface
     int SetGateOpen(bool open);
@@ -458,6 +502,7 @@ private:
     bool open_;
     long pos_;
     long numPos_;
+    long highestLabeledPos_;
     uInt32 portWidth_;
     long inputLine_;
     long firstStateSlider_;
--- a/DeviceAdapters/NIDAQ/NIDAQ.h
+++ b/DeviceAdapters/NIDAQ/NIDAQ.h
@@ -424,6 +424,10 @@ public:
 
     virtual unsigned long GetNumberOfPositions()const { return numPos_; } // replace with numPos_ once API allows not creating Labels
 
+    // Gate (shutter) interface
+    int SetGateOpen(bool open);
+    int GetGateOpen(bool& open);
+
     // action interface
     // ----------------
     int OnState(MM::PropertyBase* pProp, MM::ActionType eAct);
--- a/DeviceAdapters/NIDAQ/NIDigitalOutputPort.cpp
+++ b/DeviceAdapters/NIDAQ/NIDigitalOutputPort.cpp
@@ -35,6 +35,7 @@ DigitalOutputPort::DigitalOutputPort(const std::string& port) :
    open_(true),
    pos_(0),
    numPos_(0),
+   highestLabeledPos_(-1),
    portWidth_(0),
    nrOfStateSliders_(4),
    inputLine_(8),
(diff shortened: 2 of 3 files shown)


alandolt:
I just attempted to set up the new adapter on a second microscope (both tests were done using an Ni-6353 connected over PCIe). During this process, I discovered that the commit introducing the label property can, at least in some cases, lead to an error during the initial hardware setup, as the initial labels cannot be generated.

However, this test was performed on a system running a very old NI-MAX version (from 2018). Therefore, the entire merge request requires additional validation. I will perform further testing next week.

Sorry for the inconveniences caused.

marktsuchida:
Thanks, @alandolt! What you describe is a very valid use case.

@nicost Can you review the NIDAQ changes? Was the Label property left out intentionally, by any chance?

I didn't spot any behavioral issues in the gating support, but it's a little fragile because the base class has its own `geteOpen_` that is not kept in sync with the newly added `open_`. Perhaps `SetGateOpen()` could delegate to the base class in the normal case: something like

```c++
  int DigitalOutputPort::SetGateOpen(bool open)
  {                 
     if (blanking_ || sequenceRunning_)
     {              
        open_ = open;                   
        return DEVICE_OK;            
     }
     open_ = open;
     return CStateBase::SetGateOpen(open);                                      
  }
```

You then don't need to override `GetGateOpen()`.

---

As for the PVCAM changes (Cc: @tomhanak):

> the PVCAM adapter was updated to allow explicitly switching off the light source. In addition, abortAcquisitionInternal in PVCAMUniversal.cpp now notifies the core system when an acquisition finishes

I see the changes for the latter (2 added calls to `AcqFinished()`) but no changes that look like the former — is that right?

More importantly, in polling mode there is already a call to `AcqFinished()` in `PollingThreadExiting`, which should have run before `pollingThd_->wait()` returns. So I think the first of the two added calls to `AcqFinished()` results in a double call.

The second added call looks like a correct fix for when `CircularBufferEnabled` is turned off.

Can you give that a check?

---

PS Not a big deal for these relatively small changes, but we'd appreciate it if you could create separate pull requests for separate changes (NIDAQ vs PVCAM in this case). It makes it easier if we want to accept one change but not the other, for example. Again, no worries for this time, and thanks for keeping them in separate commits — we much appreciate your PR!

tomhanak:
Unfortunately, maintaining correct acquisition state in all supported cases is very fragile and innocent changes like this usually break something. Without spending a day on testing I cannot clearly say what happens. Simplifying the code has been on my TODO list for quite some time now. 

On the other hand, this won't be versioned release so if it breaks something, people can open new issue and use older nightly builds until it gets fixed...

> More importantly, in polling mode there is already a call to AcqFinished() in PollingThreadExiting, which should have run before pollingThd_->wait() returns. So I think the first of the two added calls to AcqFinished() results in a double call.

I agree with that.

alandolt:
Hi all,
I went back to the code and did some additional testing, and I want to clarify and correct a few things.

**Hardware clarification**
I tested the device adapter on two NI DAQ devices:
- PCIe-6321 (8-bit output range, up to 256 states)
- PCIe-6323(32-bit output range)

I did not test on a PCIe-6353, sorry for the mix-up, I'm sometimes bad with numbers. 

**Behavior with the original adapter (latest nightly, from end of March)**
- PCIe-6321: Everything works fine. In the hardware configuration wizard, labels can be assigned to each state, but they don’t show up later in the Property Browser.
- PCIe-6323: The hardware configuration hangs when it reaches the label setup page. This is likely because it tries to iterate over and auto-generate labels for all possible states (probably an existing issue for all wide-state devices).

My previous commit related to labels did not fix this.

**Fix for label handling**
I revisited this and, with commit  
[cec8dfa](https://github.com/micro-manager/mmCoreAndDevices/pull/907/commits/cec8dfadd6fa11e5ed3c55943969e49e2fe6b207), this should now be resolved. Current behavior (tested on both devices):

- Normal wide output devices (≤256 states)
  Labels are shown in the hardware configuration wizard and can be set there.
- Wide output devices (>256 states)  
  No labels are shown during hardware configuration, but they can be defined in the Micro-Manager config.

In both cases, labels are exposed via device properties, so users can assign meaningful names (e.g. mapping lamps or lasers without having to remember that UV is state 1 and cyan is state 4). The hanging issue on the PCIe-6323 (and other wide-state devices) should now also be fixed.

**Other changes**
Thanks for the feedback on method overrides @marktsuchida . I removed the `GetGateOpen` override and a few related things in  
[2277a54](https://github.com/micro-manager/mmCoreAndDevices/pull/907/commits/2277a5448a37cbf8abf8fe762565f2eade5af9a5).

Regarding the double call to `AcqFinished`: you were right @tomhanak . This is now removed in  
[5f81cbc](https://github.com/micro-manager/mmCoreAndDevices/pull/907/commits/5f81cbc0601f0b75c27a1d8c513ba48c3ea9692d).

**Testing**
I did a quick test on both setups:
- PCIe-6321 with a Teledyne Prime BSI Express
- PCIe-6323 with an Andor Zyla 4.2 Plus

Both behave as expected: labels and state can be selected correctly, and when blanking is set to off, camera synchronization is disabled.

Thanks also for the hint about splitting things into multiple PRs. I’ll do that in the future. Sorry for the size of this one; it’s my first bigger contribution to mmCoreAndDevices (my changes in the utility adapter for the PropertyShutter were much smaller 🙂).

nicost:
> @nicost Can you review the NIDAQ changes? Was the Label property left out intentionally, by any chance

If I remember correctly, I did not add labels to prevent 32,000 lines to appear in the configuration file.  There may be ways to have labels and avoid adding them all to the config file (there even is a PR open for this issue), but it seemed simpler to me to name things at a higher level.  Will look at the code soon!

alandolt:
@nicost this did not happen in my tests. I think this behaviour had been generally fixed in uManager (wasn't there a commit that changed this behaviour so that only populated fields will go into the config file). 
Is there anything else I should test, or do you (@nicost @tomhanak @marktsuchida) need anything else from me?

nicost:
This does add a lot of lines with unused labels in the configuration files (I just tested with NIMAX emulated NIDAQ boards).  I do not think we should add this until the issue with labels that are not actually used is resolved (#680).  Could you create a new PR with only your initial changes (omitting the Label additions) to the NIDAQ and another PR for the single change to the PVCAM adapter?  Sorry for the extra work, but that would make things much cleaner and easier to oversee.  

B.t.w., I see that you are working of your main branch, it is invariably a good idea to make a new branch to work on something, and keep your main branch in sync with the upstream main branch.  Not quite sure how you can do that now without nuking things, maybe @marktsuchida has suggestions.

nicost:
Finally taking some time to look at this carefully (very sorry it took so long!).  However, I am a bit surprised to find that the GateOpen functionality is already implemented.  The current code (in the main branch) uses the Get and SetOpen functions inherited from CStateDevice.  The OnState function uses the following code (which looks correct and functions correctly in my limited testing):

```
bool gateOpen;
GetGateOpen(gateOpen);
long pos;
pProp->Get(pos);
if ((pos == pos_) && (open_ == gateOpen))
   return DEVICE_OK;

long closed_state;
GetProperty(MM::g_Keyword_Closed_Position, closed_state);
long newState = gateOpen ? pos : closed_state;

// pause blanking, otherwise most cards will error
int err;
if (blanking_)
{
   err = GetHub()->StopDOBlankingAndSequence(portWidth_);
   if (err == DEVICE_OK)
      err = GetHub()->StartDOBlankingAndOrSequence(niPort_, portWidth_, true, false, newState, blankOnLow_, triggerTerminal_);
}
else
{
   err = SetState(newState);
}
```

In combination with a State Device Shutter this works as intended.  So, can you elaborate a bit more on 

> Previously, the NIDAQ adapter did not function correctly when hardware blanking was set to off. Opening or closing the shutter failed to drive the digital output back to 0 because the internal logic implicitly relied on blanking"

I would rather fix the existing code than add a second code path intended to do the same thing.  However, I would first need to understand what exactly does not work in the current code.

As for the PVCAM change, PR #929 cleans up the PVCAM code quite a bit.  If you feel the additional line of code is still needed there, can you either bring it up in that PR, or submit a new PR after #929 is merged and explain under which circumstances it is needed (I find it difficult to understand which situations trigger the code path in which that lines woudl be executed).

tomhanak:
@alandolt I've recently started with code cleanup in PVCAM adapter and among other changes I'm considering complete removal of polling mechanism (doesn't work relyably with current fast cameras). The adapter would still support acquisitions with circular buffer or a loop of single snaps, but both with callbacks only. I would like to better understand your usage of Photometrics cameras before removing something you depend on...

In your first commit to PVCAM adapter in this PR you wrote:
>Add AcqFinished() calls to the two missing paths, matching the existing behavior of the callbacks-enabled path. Usefull in combination with the NIDAQ patch when you want to use a NIDAQ without using blanking.

The `AcquisitionMethod` property is set to `Callbacks` by default. Did you change it intentionally to `Polling` (why?) or was it changed indirectly by some component (e.g. NIDAQ adapter, MM via Multi D. Acquisition, etc.)?

According to my testing the auto-shutter doesn't close (guessing based on icon in main window) only when `CircularBufferEnabled` property is set to `OFF` which is also non-default value. Did you change that property? Here it could happen internally e.g. in Multi D. Acquisition with non-zero interval between frames, because PVCAM adapter must switch from circ. buffer mode to single snaps...

nicost:
Since the functionality was already in the NIDAQ adapter (I'll be happy to look into reasons this did not work for you) and the PVCAM adapter already has changed, is it OK to close this PR?

alandolt:
@nicost I will re-test this on my setup using a completely fresh installation with the latest PVCAM and NI-DAQ adapters. Regarding configuration pollution, my observations differ: in tests on real hardware, I did not encounter this issue.

@tomhanak I did not explicitly set the startup configuration to Polling; I need to verify what the current setting is. You are, however, correct that I set CircularBufferEnabled to Off. When using pymmcore-plus (which I use to control the microscope, with pipelines built on top of it and napari-micromanager as the GUI), enabling the circular buffer causes a crash the first time an MDA is launched via run_mda because the circular buffer is not initialized. This issue does not occur when running the same MDA via the Micro-Manager GUI. We partially mitigated this by disabling the circular buffer, although I need to re-check the details, this also required a partially custom MDA implementation to work around related issues.

Given the current state, let's as you have suggested close this pull request for now. Once I have fully re-tested everything on my setup using the most recent versions, we can revisit the discussion. Thank you, @nicost, for the offer to help, I will certainly follow up. Also, please note that this setup is not mine, so I do not have daily access for testing.

PR #912, opened 2026-04-06
ThorlabsUSBCamera: signal event after StopSequenceAcquisition

tlambert03:
when running two thorlabs USB cams via MultiCam, it takes a very long time to stop.  Debugging indicated that the wait time was after MultiCam sent the request to thorlabs, and setting a stop event was suggested (by claude) as the best way to fix the problem).


tlambert03:
was able to compile and test on a real device and it fixed the issue I observed

PR #926, opened 2026-04-26
Rename unimplemented intervalMs param to unused

marktsuchida:
The `intervalMs`/`interval_ms` parameter to `CMMCore::startSequenceAcquisition()` and `MM::Camera::StartSequenceAcquisition()` are not implemented in the vast majority of cameras. And frame rate control should work the same way as exposure or gain control, that is, it should be a property (or a dedicated function, or both). It doesn't make sense to have a special place for frame rate/interval only.

There is no plan to make this work for all (or most) cameras, so let's present it honestly: it's an unused parameter.

Except, of course, when its not unused:

- `Andor` (not `AndorSDK3`) does set the frame interval using this parameter (and does not have a corresponding property).
- `HamamatsuHam` appears to set up the acquisition slightly differently when interval_ms is > 0, although it does not use the value.

So, at least for now, we keep the plumbing and pass the `unused` parameter all the way to the camera device, so that behavior does not change. (Perhaps those 2 camera adapters can be updated in the future so that we can fully drop this parameter.)

### DIFF
--- a/MMCore/MMCore.cpp
+++ b/MMCore/MMCore.cpp
@@ -3022,11 +3022,17 @@ long CMMCore::getImageBufferSize()
  * Starts streaming camera sequence acquisition.
  * This command does not block the calling thread for the duration of the acquisition.
  *
- * @param numImages        Number of images requested from the camera
- * @param intervalMs       The interval between images, currently only supported by Andor cameras
- * @param stopOnOverflow   whether or not the camera stops acquiring when the circular buffer is full
+ * @param numImages       Number of images requested from the camera.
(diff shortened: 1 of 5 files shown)


nicost:
Looks good to me!

marktsuchida:
@tlambert03 This will change a param name, which could affect the Python API (if called using kwarg).

I suspect we can deal with it downstream (whether we decide to keep or rename the Python param name), but let me know if you expect special care to be needed.

tlambert03:
It's fine.
Internally, I don't think we ever use kwargs for core calls.  We'll deal with it if so

PR #929, opened 2026-04-28
PVCAM: Code cleanup and modernization

tomhanak:
Although the amount of changes looks huge, there are no functional or logical changes. The only exception is a fix in used bayer mask constant (`CFA_RGGB` instead of `COLOR_RGGB`) in a code branch that should never happen.

For review please follow the commits where each is focused to one simple change.


nicost:
Thanks!  All looks very clean to me.  Happy to learn from your code that I can now initialize variables in the header file, that makes so much more sense!  @marktsuchida  OK to merge?

DeviceAdapters/PVCAM/PVCAMAdapter.h
```diff
@@ -834,114 +814,112 @@ class Universal : public CCameraBase<Universal>
...
-    bool            circBufFrameRecoveryEnabled_; // True if we perform recovery from lost callbacks
+    PvCameraModel   cameraModel_{ PvCameraModel_Generic };
+    char            deviceLabel_[MM::MaxStrLength]{ '\0' }; // Cached device label used when inserting metadata
 
-    bool            stopOnOverflow_;       // Stop inserting images to MM buffer if it's full
-    bool            snappingSingleFrame_;  // Single frame mode acquisition ongoing
-    bool            singleFrameModeReady_; // Single frame mode acquisition prepared
-    bool            sequenceModeReady_;    // Continuous acquisition prepared
-    bool            callPrepareForAcq_;    // Call PrepareForAcq after {sequence,singleFrame}ModeReady_ is set
+    int             circBufFrameCount_{ 10 }; // number of frames to allocate the buffer for
```
marktsuchida, line 843: Should this be `CIRC_BUF_FRAME_CNT_DEF` (== 20)? (Also set in constructor; might be good to set in one place.)

marktsuchida (commented):
Nice modernization, thanks!

One minor comment which you can optionally address, but looks ready to merge.

DeviceAdapters/PVCAM/PVCAMAdapter.h
tomhanak: Yes, but I would like to keep this as it is. Not initializing this non-constant variable in header would look inconsistent and moving `CIRC_BUF_FRAME_CNT_DEF` to header would unnecessary expose these constants to other classes.
I was considering an invalid value (like 0 or -1) or some sane default that differs from the actual value known in `.cpp` file only...

marktsuchida: Got it, thanks!

PR #933, opened 2026-05-01
Fix Sapphire laser power readback

tzhaoo:
The issue was reported on the forum: https://forum.image.sc/t/power-setting-readback-issue-sapphire-laser/115005?u=evdwee. I encountered the same issue when changing the laser power in Micro-Manager, the actual laser output changes correctly, but the displayed `PowerSetpoint` and `PowerReadback` values can be incorrect immediately after setting the power. For example, if the current power is 20 mW and the slider is moved to 50 mW, the laser output reaches the requested power, but the displayed value may remain close to the old output value, such as around 23 mW. Refreshing the device properties later shows the correct value.

The main cause was that `setLaser()` sent `P=<value>` and then immediately queried `?P`. Since `?P` reports actual output power, this could return a transient value while the laser was still moving toward the new setpoint. The adapter then used that transient value as the achieved PowerSetpoint. The [Sapphire OEM manual](https://acmerevival.com/wp-content/uploads/2022/10/coherent-sapphire-488-operator-s-manual-84.pdf) distinguishes `?P`, actual output power, from `?SP`, configured power setpoint. This PR uses `?SP` for `PowerSetpoint`, keeps `?P` for `PowerReadback`, and sends `P=<value>` directly when setting laser power without expecting an echo or immediately querying `?P`.

Another cause of inaccuracy is that MMStudio's property browser may refresh after `PowerSetpoint` is changed, while the laser output is still changing toward the new setpoint. This is especially noticeable after large power changes. So I add a short 700 ms delay after setting the power so that the laser has time to settle closer to the requested output.

I also add additional device information properties, including head ID, usage hours, laser diode current, baseplate temperature, maximum laser power, and wavelength.

Tested with a Coherent Sapphire 488 CW 100 mW laser and Micro-Manager_20260430. And I am happy to do some new test.

### DIFF
--- a/DeviceAdapters/Sapphire/Sapphire.cpp
+++ b/DeviceAdapters/Sapphire/Sapphire.cpp
@@ -87,15 +87,18 @@ Sapphire::Sapphire(const char* name) :
    busy_(false),
    changedTime_(0.0),
 	queryToken_("?"),
-	powerSetpointToken_("P"),
+	powerSetpointToken_("SP"),
 	powerReadbackToken_("P"),
 	//CDRHToken_("CDRH"),  // if this is on, laser delays 5 SEC before turning on
    //CWToken_("CW"),
 	laserOnToken_("L"),
 	TECServoToken_("T"),
 	headSerialNoToken_("HID"),
-	headUsageHoursToken_("HH")
-	//wavelengthToken_("WAVE"),
+	headUsageHoursToken_("HH"),
+	laserDiodeCurrentToken_("C"),
+	basePlateTemperatureToken_("BT"),
+	maximumLaserPowerToken_("MAXLP"),
+	wavelengthToken_("WAVE")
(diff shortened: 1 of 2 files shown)


nicost:
Check failed because the Ubuntu servers are down;).

Changes look good to me, very difficult to be sure without testing on actual hardware, but you definitely have done so.  I also can not oversee if there are issues with different version of Sapphire firmware, but I am sure we will eventually hear about that if there are issues.  

Small request, we try to use 3 spaces rather than tabs everywhere (but I believe that Mark would like to change that to 4).  This file has mixes usage of tabs and 3 spaces.  In the future, it would be nice to have 3 spaces everywhere (do not worry about that now).

tzhaoo:
Thanks for merging! We have 2 Sapphire lasers here, so we're happy to do some testing if any issues come up. And I'll make sure to follow the style conventions going forward.

PR #962, opened 2026-06-19
SpinnakerC: Fix bits-per-pixel query

marktsuchida:
Likely fix for https://forum.image.sc/t/micro-manager-and-spinnakercamera/64076/17

Avoid the enum int value (which can depend on device) and use the symbolic name, as was done in the original Spinnaker adapter.

- [x] Test on Windows that it builds and works


marktsuchida:
Works with a BFS-U3-16S2M (which also worked before this fix).

PR #969, opened 2026-07-08
Add PicardUSB device adapter

marktsuchida:
Picard Industries devices using their PiUsbSDK.

For now, just supports USB Gradient Wheel and USB Side Flipper.

- [x] Test side flipper
- [ ] Side flipper Busy() - add extra delay?
- [ ] Test gradient wheel
- [ ] Gradient wheel Busy() - add extra delay?
- [ ] SDK must be in 3rdpartypublic
- [ ] SDK DLL conflicts with PicardStage adapter; decide what to do (probably just ship new one)

### DIFF
--- a/DeviceAdapters/PicardUSB/PicardUSB.cpp
+++ b/DeviceAdapters/PicardUSB/PicardUSB.cpp
@@ -1,0 +1,383 @@
+// Picard Industries USB devices (PiUsbSDK)
+//
+// AUTHOR:        Mark A. Tsuchida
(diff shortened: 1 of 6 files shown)


marktsuchida:
Might need to revisit the choice of making the Side Flipper a Shutter device, because the current shutter is closed on shutdown (`unloadAllDevices()` or `CMMCore` destructor). That's usually safe for shutters, but an arm swinging when closing MM may not be great.

PR #970, opened 2026-07-14
PVCAM: Fixed regression with shutter

tomhanak:
The `PrepareForAcq` callback must be called unconditionally to properly notify core whenever new acquisition is about to start.


nicost:
Thanks!

PR #972, opened 2026-07-17
Correct IsMoving

NigelWagg:
Skip extra read byte to correct IsMoving


NigelWagg (commented):
Skip device byte in ScientificaMotion8.cpp
The Is_Moving method was returning the Device number instead of the moving flag, a dummy read fixes this.

marktsuchida:
Thanks!

PR #973, opened 2026-07-18
Use standard C++ for DeviceThreadBase and MMThreadLock implementation

### LINKED ISSUE #971: `MMDeviceThreadBase` leaks Win32 thread handle
https://github.com/micro-manager/mmCoreAndDevices/blob/61567cf33c60426f87663174368d86dee0fd1985/MMDevice/DeviceThreads.h#L33-L59

The [`CreateThread` docs](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createthread) are clear that the return handle (if not null) must be closed.

Closing the handle neither joins nor terminates the thread, so we probably just need to close it in the destructor.

There is a similar problem with pthreads: we do not detach the thread, so it remains joinable; if we don't join, then the kernel keeps a record of the thread until the process exits. We should detach in the destructor (when not already joined) to plug this leak without breaking compatibility.

As always, new device code should always use `std::thread` (`std::jthread` once we move to C++20) instead.

marktsuchida:
Closes #971.

Retire the Win32 and pthreads types/calls and replace with `std::thread` and `std::recursive_mutex`.

No longer leak a thread handle (Windows) or joinable thread (POSIX).

Patch device adapters that broke because MMDevice no longer includes Windows.h or defines `WIN32_LEAN_AND_MEAN`.

No device ABI changes. Source-incompatible changes (no in-tree adapters affected):
- `DeviceThreadBase::activate()` is no longer `virtual` and now returns `void` rather than `int` (no adapter overrides this function or reads its return value)
- `MMThreadGuard::isLocked()` has been deleted (no adapter calls it)

TODO:
- [x] Patch device adapters (if any) that break because pthread.h is no longer included by MMDevice

### DIFF
--- a/DeviceAdapters/89NorthLDI/LDI.cpp
+++ b/DeviceAdapters/89NorthLDI/LDI.cpp
@@ -2,6 +2,9 @@
 
 #include <ctime>
 
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+
 const char* g_LDI_name = "89 North Laser Diode Illuminator";
 const char* g_LDI_description = "Multi-line, Solid-State Laser Illuminator";
 #define LDI_ERROR 108901
--- a/DeviceAdapters/ABS/AbsImgBuffer.h
+++ b/DeviceAdapters/ABS/AbsImgBuffer.h
@@ -2,6 +2,9 @@
 #include "ImgBuffer.h"              //!< base  class
 #include "DeviceThreads.h"          //!< MMThreadLock class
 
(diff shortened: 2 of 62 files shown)


marktsuchida:
There were no missing `pthread.h` includes in device adapters, as far as I can easily test, on macOS (official build) and Linux (best-effort). Not surprising because (unlike Windows.h) it's only needed for threading. Also, libstdc++/libc++ appear to leak `pthread.h` from `<thread>` and others, so we likely cannot break existing code by this `#include` removal in practice.

PR #975, opened 2026-07-27
Fix/aravis

HazenBabcock:
This PR fixes the Aravis driver so that it will compile with the current version of micro-manager.

I also changed the supported Aravis version to 0.8 as this is the stable version and is hopefully easier to install. It is for example available in most linux package managers.


marktsuchida:
Thanks!

PR #981, opened 2026-08-13
Aravis: fix memory safety bugs, including a crash on any camera with a fixed ROI

HazenBabcock:
## AI Tools Note

The following summary was generated by Claude Code. I have reviewed it's summary 
and the changes it proposed and believe them to be correct.

## Summary

Six memory-safety fixes in the Aravis camera adapter. One is a crash that stops
any camera whose ROI cannot be set from opening at all.

No new features, and no intended behaviour change beyond not crashing — the
pixel formats recognised, the ROI handling and the property set are all as
before. Correctness fixes (`GetROI`, bit depth reporting, RGB byte order) and
capability handling are deliberately left for follow-up changes to keep this one
reviewable.

## The crash

`ArvCheckError()` took its `GError *` by value:

```c
int AravisCamera::ArvCheckError(GError *gerror) const
{
  if (gerror != NULL) {
    std::stringstream msg;
    msg << "Aravis Error: " << gerror->message;
    LogMessage(msg.str(), false);
    g_clear_error(&gerror);      // frees the error, clears only the local copy
    return 1;
  }
  return 0;
}
```

`g_clear_error()` frees the error and NULLs the pointer it is given — but that
is this function's own copy. The caller's pointer still points at freed memory,
and callers reuse one `GError *` across several Aravis calls:

```c
int AravisCamera::ClearROI()
{
  GError *gerror = nullptr;

  arv_camera_set_region(arv_cam, 0, 0, 64, 64, &gerror);   // fails -> allocates GError
  ArvCheckError(gerror);                                    // logs, frees, caller dangles

  arv_camera_get_height_bounds(arv_cam, &tmp, &h, &gerror); // *error is not NULL
  ArvCheckError(gerror);                                    // reads freed memory
```

`Initialize()` calls `ClearROI()`, whose first action is `arv_camera_set_region()`.
On a camera that does not allow the region to be set that call fails, and the
next check dereferences the freed error. The result is a segfault while opening
the camera.

This is why it has gone unnoticed: on a camera with an adjustable ROI the
`set_region()` succeeds, no error is ever allocated, and the path is never
taken.

Under AddressSanitizer:

```
SUMMARY: AddressSanitizer: heap-use-after-free
         AravisCamera.cpp:281 in AravisCamera::ArvCheckError(_GError*) const
   freed by: g_set_error <- arv_camera_set_region <- ClearROI:426 <- Initialize:554
```

`ArvCheckError()` now takes a `GError **`. That is the bulk of the diff: 70 call
sites, mechanically changed to pass `&gerror`.

## Also in this PR

- **The camera name was copied into a buffer one byte too small.** The
  constructor did `malloc(strlen(name))` and then
  `CDeviceUtils::CopyLimitedString()`, which writes `strlen(name) + 1` bytes, so
  the terminating NUL landed past the end of the allocation for every camera.
  The name is now a `std::string`.

- **The stream's buffer pop was a side effect of `g_assert()`.** Built with
  `G_DISABLE_ASSERT` the pop would disappear and the stream would starve once
  its buffers ran out; a genuine mismatch aborted the application rather than
  being handled. The pop is now unconditional, and a NULL or unexpected buffer
  is logged.

- **`Shutdown()` did nothing.** It now stops a running acquisition, releases the
  stream and the snap buffer, and frees the image buffer. It is idempotent and
  the destructor calls it, so unloading a configuration no longer leaks or
  leaves a stream writing into a freed buffer.

- **`capturing` was a plain `bool`** written by the Micro-Manager thread and
  read by the Aravis stream callback thread. It is now `std::atomic<bool>`, and
  a mutex guards the image buffer, which the callback may reallocate while
  `GetImageBuffer()` is reading it.

- **`SnapImage()` could hang forever.** It passed a zero timeout to
  `arv_camera_acquisition()`, which Aravis documents as "no timeout" and
  implements with a blocking pop. A camera waiting on a hardware trigger, or a
  single dropped packet, blocked the calling thread with no way out. It now
  waits five times the exposure time with a five second floor, and logs on
  timeout.

- **An unimplemented pixel format overflowed the image buffer.**
  `ArvPixelFormatUpdate()`'s `default` branch returned without touching the
  image description, leaving the constructor's zeros on a camera whose format is
  unsupported from the start. Zero components is not one, so `ArvBufferUpdate()`
  took the packed-RGB path and wrote four bytes per pixel into a buffer
  allocated for `width * height * 0`. The default branch now zeroes the
  description explicitly and logs to the Micro-Manager log rather than stdout,
  and `ArvBufferUpdate()` refuses to copy when the description is unusable or
  when the source buffer is smaller than the copy would read.

## Testing

Linux, Aravis 0.8.36, against a GigE Vision emulator over loopback presenting
five capability profiles: a minimal camera (exposure only, fixed geometry), a
mono camera (Mono8/10/12/16), a feature-rich camera (gain, black level, gamma,
auto modes, trigger, binning, temperature), a colour camera (Bayer, RGB8, BGR8),
and a camera whose current format the adapter does not implement.

| profile | before | after |
|---|---|---|
| minimal | SIGSEGV in `Initialize` | opens, snaps, streams |
| mono | SIGSEGV in `Initialize` | opens, snaps, streams |
| rich | SIGSEGV in `Initialize` | opens, snaps, streams |
| colour | SIGSEGV in `Initialize` | opens, snaps, streams |
| unsupported | SIGSEGV in `Initialize` | fails cleanly with a logged reason |

With the adapter built `-fsanitize=address`: no use-after-free, no
heap-buffer-overflow, and no leak report naming the adapter. Ten
load / initialize / snap / unload cycles grow RSS by 0 kB, which covers both
that `Shutdown()` releases and that it tolerates running twice.

### DIFF
--- a/DeviceAdapters/Aravis/AravisCamera.cpp
+++ b/DeviceAdapters/Aravis/AravisCamera.cpp
@@ -275,6 +275,17 @@ void AravisCamera::ArvBufferUpdate(ArvBuffer *aBuffer)
   arvPixelFormat = arv_buffer_get_image_pixel_format(aBuffer);
   ArvPixelFormatUpdate(arvPixelFormat);
 
+  // A format with no case in ArvPixelFormatUpdate() leaves these at zero.
+  // Zero components is not one, so the copy below would take the RGB path and
+  // write four bytes per pixel into a buffer sized for zero. Refuse instead.
+  if ((img_buffer_bytes_per_pixel < 1) || (img_buffer_number_components < 1)){
+    std::stringstream msg;
+    msg << "Aravis Error, cannot copy an image in unsupported pixel format "
+	<< arvPixelFormat;
+    LogMessage(msg.str(), false);
+    return;
+  }
+
   // Image size updates.
   img_buffer_width = (int)arv_buffer_get_image_width(aBuffer);
   img_buffer_height = (int)arv_buffer_get_image_height(aBuffer);
@@ -283,6 +294,21 @@ void AravisCamera::ArvBufferUpdate(ArvBuffer *aBuffer)
   // Copy buffer to MM.
   arvBufferData = (unsigned char *)arv_buffer_get_data(aBuffer, &arvSize);
   size = img_buffer_width * img_buffer_height * img_buffer_bytes_per_pixel;
+
+  // The source must hold everything the destination is about to be filled
+  // with. For the packed RGB formats the camera sends three bytes per pixel
+  // and we expand to four, so compare against what will actually be read.
+  size_t arvNeeded = (img_buffer_number_components == 1)
+    ? size
+    : img_buffer_number_pixels * 3;
+  if (arvSize < arvNeeded){
+    std::stringstream msg;
+    msg << "Aravis Error, buffer holds " << arvSize << " bytes but "
+	<< arvNeeded << " are needed for a " << img_buffer_width << "x"
+	<< img_buffer_height << " image";
+    LogMessage(msg.str(), false);
+    return;
+  }
 
   if (img_buffer_size != size){
     if (img_buffer != nullptr){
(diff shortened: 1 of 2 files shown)


marktsuchida:
Thanks! Sorry for the wait.

PR #989, opened 2026-09-14
Add ClearPropertyLimits, use it in PVCAM to stop spurious log messages

ieivanov:
## Problem

There is currently no way for a device adapter to remove a property's limits without the Core logging an error. The only idiom available is `SetPropertyLimits(prop, 0, 0)`: `MM::Property::SetLimits()` does clear the limits for an empty range, but it returns `false`, and `CDeviceBase::SetPropertyLimits()` treats that as a failure — it returns `DEVICE_INVALID_PROPERTY_LIMTS` and logs:

```
Device adapter requests invalid values ( 0, 0) for property: <name>
```

PVCAM uses this idiom every time a property is made read-only, so the message shows up repeatedly during normal operation — once per acquisition setup, which in a long MDA means once per event:

```
[IFO,dev:Prime BSI Express] Device adapter requests invalid values ( 0, 0) for property: CircularBufferFrameCount
```

The PVCAM source already documented this as a known wart and skipped error checking at those call sites because of it.

## Change

- `MM::Property::ClearLimits()` — clears `limits_` and resets the stored bounds.
- `CDeviceBase::ClearPropertyLimits(name)` — returns `DEVICE_OK` and logs nothing; returns `DEVICE_INVALID_PROPERTY` if the property does not exist.
- PVCAM: the four intentional limit-clearing sites now call `ClearPropertyLimits` — `CircularBufferFrameCount` (in `Initialize()` and `updateCircBufRange()`), `ScanWidth`, and `ScanLineDelay`. Obsolete comments about the workaround removed.

`SetPropertyLimits()` behavior is unchanged, so genuinely invalid ranges (`low >= high`, or limits on a property with discrete allowed values) are still reported as before.

## Testing

`MMDevice/Property.cpp` compiles clean, and `ClearPropertyLimits` was instantiated in a throwaway `CGenericBase` device to check the template. The PVCAM adapter itself was not compiled — it needs the PVCAM SDK and Windows, which I do not have here — so those four call sites are reviewed but not compiler-checked; each is a rename of an existing method call on the same class.

Other adapters (`Mightex_C_Cam`, `Mightex_SB_Cam`, `Revealer`) use the same `0, 0` idiom and would benefit from the same one-line switch; left out of this PR to keep it focused, happy to include them if preferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


marktsuchida:
Nice fix, thanks!

